PR #2, opened 2021-11-05
build: Adjust for kernel mctp.h

kraj:
Starting 5.15 kernel has mctp.h as well and provides some of the
defintions, therefore move the ones we need to provide in this header
outside the conditional check so its always provided regardless of
header coming from kernel or not.

Include linux/if_link.h so it can pre-define IFLA_MCTP_UNSPEC and other
defines and there is a check for conditionally providing them here but
kernel does not have such a check, so if linux/if_link.h is included
after this header then we end up with duplicate defines

linux/if_link.h:1267:2: error: redefinition of enumerator 'IFLA_MCTP_UNSPEC'
        IFLA_MCTP_UNSPEC,
        ^
../git/src/mctp.h:74:2: note: previous definition is here
        IFLA_MCTP_UNSPEC,
        ^

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/mctp.h
+++ b/src/mctp.h
@@ -19,6 +19,7 @@
 #if HAVE_LINUX_MCTP_H
 #include <linux/mctp.h>
 #endif
+#include <linux/if_link.h>
 
 #ifndef AF_MCTP
 #define AF_MCTP 45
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
Hi Khem,

Thanks for the PR, and for testing against the actual kernel headers! Looks good, merged.

PR #15, opened 2023-10-06
mctpd: Add `NotifyDiscovery` D-Bus method

khangng-ampere:
This method sends a Discovery Notify to the PCIe Root Complex.

Tested:

    busctl call xyz.openbmc_project.MCTP /xyz/openbmc_project/mctp \
        au.com.CodeConstruct.MCTP NotifyDiscovery s mctppcie0

should work.

However, the message may take a long time to return so the timeout may need to be increased. This is to be investigated.

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -1832,6 +1832,66 @@ err:
 	return rc;
 }
 
+static int method_notify_discovery(sd_bus_message *call, void *data,
+				   sd_bus_error *berr)
+{
+	int rc;
+	const char *ifname = NULL;
+	dest_phys desti = { 0 }, *dest = &desti;
+	ctx *ctx = data;
+	struct mctp_ctrl_cmd_discovery_notify req = { 0 };
+	struct mctp_ctrl_resp_discovery_notify *resp;
+	uint8_t *buf;
+	size_t buf_size;
+	struct sockaddr_mctp_ext resp_addr;
+
+	rc = sd_bus_message_read(call, "s", &ifname);
+	if (rc < 0)
+		goto err;
+
+	dest->hwaddr_len = 2;
+	dest->ifindex = mctp_nl_ifindex_byname(ctx->nl, ifname);
+	if (dest->ifindex <= 0)
+		return sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+					 "Unknown MCTP ifname '%s'", ifname);
+
+	rc = validate_dest_phys(ctx, dest);
+	if (rc < 0)
+		return sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+					 "Bad physaddr");
+
+	req.ctrl_hdr.command_code = MCTP_CTRL_CMD_DISCOVERY_NOTIFY;
+	req.ctrl_hdr.rq_dgram_inst = RQDI_REQ;
+
+	rc = endpoint_query_phys(ctx, dest, MCTP_CTRL_HDR_MSG_TYPE, &req,
+				 sizeof(req), &buf, &buf_size, &resp_addr);
+	if (rc < 0)
+		goto free_buf;
+
+	if (buf_size != sizeof(*resp)) {
+		warnx("%s: wrong reply length %zu bytes. dest %s", __func__,
+		      buf_size, dest_phys_tostr(dest));
+		rc = -ENOMSG;
+		goto free_buf;
+	}
+	resp = (void *)buf;
+
+	if (resp->completion_code != 0) {
+		// TODO: make this a debug message?
+		warnx("Failure completion code 0x%02x from %s",
+		      resp->completion_code, dest_phys_tostr(dest));
+		rc = -ECONNREFUSED;
+		goto free_buf;
+	}
+	free(buf);
+	return sd_bus_reply_method_return(call, "");
+free_buf:
+	free(buf);
+err:
+	set_berr(ctx, rc, berr);
+	return rc;
+}
+
 // Query various properties of a peer.
 // To be called when a new peer is discovered/assigned, once an EID is known
 // and routable.
@@ -2274,6 +2334,14 @@ static const sd_bus_vtable bus_mctpd_vtable[] = {
 		SD_BUS_PARAM(found),
 		method_learn_endpoint,
 		0),
+
+	SD_BUS_METHOD_WITH_NAMES("NotifyDiscovery",
+		"s",
+		SD_BUS_PARAM(ifname),
+		"",,
+		method_notify_discovery,
+		0),
+
 	SD_BUS_VTABLE_END,
 
 };
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -51,6 +51,29 @@ Like SetupEndpoint but will not assign EIDs, will only query endpoints for a cur
 The `new` return value is set to `false` for an already known endpoint, or `true` when an
 endpoint's EID is newly discovered.
 
+## `.NotifyDiscovery`
+
+This method is used to trigger the MCTP over PCIe-VDM discovery process on a
+PCIe-VDM interface.
+
+This method sends a Discovery Notify message to the Root Complex on the PCIe bus
+with null EID, null physical address and Route-To-Root-Complex PCIe-VDM routing
+type.
+
+This method should only be used when `mctpd` is running on an endpoint.
+
+`NotifyDiscovery <interface name>`
+
+`<interface name>` is an PCIe-VDM interface such as `mctppcie0`.
+
+
+Example:
+
+```shell
+busctl call xyz.openbmc_project.MCTP /xyz/openbmc_project/mctp \
+        au.com.CodeConstruct.MCTP NotifyDiscovery s mctppcie0
+```
+
 ## Endpoint Methods
 
 Each endpoint object has methods to configure it, with `au.com.CodeConstruct.MCTP.Endpoint`


jk-ozlabs:
The code looks fine, but why is this triggered by a dbus method? Shouldn't mctpd be in control of when the Discovery Notify messages are sent in the first place?

Even if not, this seems like the wrong abstraction to be exposing over dbus. What's the actual state/event that we want to expose in this scenario?

(also, can you share the PCIe binding driver?)

khangng-ampere:
> The code looks fine, but why is this triggered by a dbus method? Shouldn't mctpd be in control of when the Discovery Notify messages are sent in the first place?
> 
> Even if not, this seems like the wrong abstraction to be exposing over dbus. What's the actual state/event that we want to expose in this scenario?

Yeah, I agree, at the time, I wasn't thinking much about when to send the Discovery Notify messages, so I left that to the method caller to decide. I suppose we could do this inside `mctpd` anyway.

After giving it some thought, I think we need a way to know when the PCI address is gone. This seems like something that `netlink` would provide, but atm I couldn't figure out how to tell apart media types of a link.

For example, is there a way to know where a link comes from, is it a `mctp-i2c` link or `mctp-serial` link?

ThuBaNguyen:
> The code looks fine, but why is this triggered by a dbus method? Shouldn't mctpd be in control of when the Discovery Notify messages are sent in the first place?
> 
> Even if not, this seems like the wrong abstraction to be exposing over dbus. What's the actual state/event that we want to expose in this scenario?
> 
> (also, can you share the PCIe binding driver?)

The D-Bus method provides the method consumer ability to decide the time which the BO can start initialize/discovery the BMC thru PCIe medium interface. Example, when the BMC boot up with the host is On, if BMC is set as Endpoint the other service can detect the available of the host then call `Discovery Notify` to BO to setEID for BMC.
In theory, BMC does not need to do that but I don't think propose the `Discovery Notify` will cause any potential problem.

jk-ozlabs:
> In theory, BMC does not need to do that but I don't think propose the `Discovery Notify` will cause any potential problem.

My concern is allowing callers to drive details of the MCTP Control Protocol implementation, rather than having those decisions made by mctpd internally.

However, that could be matter of making this call more of a "notify the bus owner of our presence" operation rather than "send a Notify Discovery" message. For PCIe links, that would be equivalent, but other links (ie SMBus) may be a no-op (or could return an error?).

But still: DSP0238 defines the specific enumeration / discovery process, and section 6.9.1 gives specific events which would trigger the control protocol implementation to send a Notify Discovery message. I suspect that only mctpd would have full knowledge of when these events would occur. Are you proposing external components are tracking the MCTP enumeration state? Otherwise, how do they know when it is suitable to send a Notify Discovery message?

Some knowledge of the interactions to the PCIe binding driver might be helpful here. Can you share that?

> Example, when the BMC boot up with the host is On, if BMC is set as Endpoint the other service can detect the available of the host then call Discovery Notify to BO to setEID for BMC.

Right, and that is best handled by mctpd automatically sending a Notify Discovery message when the "discovered" flag is not set, as per 6.9.1.

ThuBaNguyen:
> > In theory, BMC does not need to do that but I don't think propose the `Discovery Notify` will cause any potential problem.
> 
> My concern is allowing callers to drive details of the MCTP Control Protocol implementation, rather than having those decisions made by mctpd internally.
> 
> However, that could be matter of making this call more of a "notify the bus owner of our presence" operation rather than "send a Notify Discovery" message. For PCIe links, that would be equivalent, but other links (ie SMBus) may be a no-op (or could return an error?).
> 
> But still: DSP0238 defines the specific enumeration / discovery process, and section 6.9.1 gives specific events which would trigger the control protocol implementation to send a Notify Discovery message. I suspect that only mctpd would have full knowledge of when these events would occur. Are you proposing external components are tracking the MCTP enumeration state? Otherwise, how do they know when it is suitable to send a Notify Discovery message?

I agree with above point, maybe we don't need public the `Notify Discovery` method to D-Bus.

> 
> Some knowledge of the interactions to the PCIe binding driver might be helpful here. Can you share that?

I'm asking my manager for this. Currently, we do internal review to make the clean code before create MR to kernel.

> 
> > Example, when the BMC boot up with the host is On, if BMC is set as Endpoint the other service can detect the available of the host then call Discovery Notify to BO to setEID for BMC.
> 
> Right, and that is best handled by mctpd automatically sending a Notify Discovery message when the "discovered" flag is not set, as per 6.9.1.

Agree.

So We don't need support `Notify Discovery` to D-Bus method.

khangng-ampere:
Piggyback on recent changes, I think the most suitable time to send Discovery Notify now is when user set link role to Endpoint. I updated my PR.

jk-ozlabs:
Nice one. I'll take a look shortly.

khangng-ampere:
Hi @jk-ozlabs, do you have any concerns regarding this?

jk-ozlabs:
Ah, no concerns, just travelling at the moment. I'll look at merging when I'm back at the deck next week.

I have a v2.0 release queued though, any preference on whether this is merged before or after that?

khangng-ampere:
> Ah, no concerns, just travelling at the moment. I'll look at merging when I'm back at the deck next week.
> 
> I have a v2.0 release queued though, any preference on whether this is merged before or after that?

Let's do it after. I have a few more PRs to support the rest of the discovery process for endpoints, so they should probably be batched in one release.

src/mctpd.c
```diff
@@ -2893,6 +2893,49 @@ static int bus_link_get_prop(sd_bus *bus,
...
+	dest->ifindex = ifindex;
+	assert(mctp_nl_ifaddr_byindex(ctx->nl, dest->ifindex,
+				      &dest->hwaddr_len));
+	memset(dest->hwaddr, 0, sizeof dest->hwaddr);
+
+	req.ctrl_hdr.command_code = MCTP_CTRL_CMD_DISCOVERY_NOTIFY;
+	req.ctrl_hdr.rq_dgram_inst = RQDI_REQ;
+
+	rc = endpoint_query_phys(ctx, dest, MCTP_CTRL_HDR_MSG_TYPE, &req,
+				 sizeof(req), &buf, &buf_size, &resp_addr);
```
jk-ozlabs, line 2915: `endpoint_query_phys` will wait for a response, but Discovery Notify does not send a response. We'll probably need a different helper for datagram-style messages.

src/mctpd.c
```diff
@@ -2954,6 +3000,16 @@ static int bus_link_set_prop(sd_bus *bus,
 	}
 	lmUserData->role = role.role;
 
+	// Announce on the bus we are endpoint, print warning and ignore error if failed
+	if (lmUserData->role == ENDPOINT_ROLE_ENDPOINT) {
+		rc = notify_discovery(ctx, ifindex);
```
jk-ozlabs, line 3005: This is done unconditionally, but Discovery Notify messages should only be sent on transport types that require them. We may need a way to query the transport type?

src/mctpd.c
```diff
@@ -2893,6 +2893,49 @@ static int bus_link_get_prop(sd_bus *bus,
...
+	int rc = 0;
+	dest_phys desti = { 0 }, *dest = &desti;
+	struct mctp_ctrl_cmd_discovery_notify req = { 0 };
+	struct mctp_ctrl_resp_discovery_notify *resp;
+	uint8_t *buf;
+	size_t buf_size;
+	struct sockaddr_mctp_ext resp_addr;
+
+	dest->ifindex = ifindex;
+	assert(mctp_nl_ifaddr_byindex(ctx->nl, dest->ifindex,
```
jk-ozlabs, line 2907: this will cause bugs with `NDEBUG`. Please split the assert from the call.

src/mctp-netlink.c
```diff
@@ -722,8 +725,17 @@ static int parse_getlink_dump(mctp_nl *nl, struct nlmsghdr *nlh, uint32_t len)
 			continue;
 		}
 		ifname_len = strnlen(ifname, ifname_len);
+
+		ifaddr = mctp_get_rtnlmsg_attr(IFLA_ADDRESS, rta, rlen,
+					       &ifaddr_len);
+		if (!ifaddr) {
+			warnx("no ifaddr?");
```
jk-ozlabs, line 732: This is not a warning, nor should we skip setting up the link - some links may not have a physical address.

There are two possible (valid) behaviours there - that there is no `IFLA_ADDRESS` attribute, or that the length is reported as zero. We should allow both of those.

src/mctp-netlink.c
```diff
@@ -1067,6 +1091,12 @@ static int linkmap_add_entry(mctp_nl *nl, struct ifinfomsg *info,
 		return -1;
 	}
 
+	if (ifaddr_len > MAX_ADDR_LEN) {
+		warnx("linkmap, too long ifaddr '%*s'", (int)ifaddr_len,
+		      ifaddr);
```
jk-ozlabs, line 1096: `ifaddr` is not a string

src/mctp-netlink.c
```diff
@@ -1088,6 +1118,8 @@ static int linkmap_add_entry(mctp_nl *nl, struct ifinfomsg *info,
 	entry = &nl->linkmap[idx];
 	memset(entry, 0, sizeof(*entry));
 	snprintf(entry->ifname, IFNAMSIZ, "%*s", (int)ifname_len, ifname);
+	snprintf(entry->ifaddr, MAX_ADDR_LEN, "%*s", (int)ifaddr_len, ifaddr);
```
jk-ozlabs, line 1121: not a string here either, use memcpy.

src/mctp-netlink.c
```diff
@@ -23,6 +23,8 @@
 struct linkmap_entry {
 	int	ifindex;
 	char	ifname[IFNAMSIZ+1];
+	char	ifaddr[MAX_ADDR_LEN];
```
jk-ozlabs, line 26: probably best that this is unsigned, and so `uint8_t` instead.

jk-ozlabs (changes requested):

src/mctp-netlink.c
khangng-ampere: I wondered if I should print a hex representation of the address here, but I think as opposed to the text representation of `ifname`, the length for the address would probably be enough to track any issues.

src/mctp-netlink.c
khangng-ampere: Makes sense. Code that uses `ifaddr` should have already taken `ifaddr_len` into account (like `memcpy`), so I will remove this check.

src/mctpd.c
khangng-ampere: Since we allowed for `0` address length, this assertion should not be necessary anymore.

src/mctpd.c
khangng-ampere: @jk-ozlabs It would be great if we have that now (we need them to get control message timing information for the recovery mechanism on each interface anyway).

I would propose adding such information to when a binding driver register itself to the MCTP networking subsystem inside the kernel (`mctp_netdev_ops`), like so:

```c
static const struct mctp_netdev_ops mctp_i2c_mctp_ops = {
	.transport = MCTP_TRANSPORT_SMBUS,
	.release_flow = mctp_i2c_release_flow,
};
```

And then exposing them to userspace through netlink.

I don't know how long it would take to merge the patch to linux, but in the meanwhile, we may temporarily provide such information in `mctpd` by basing on the interface name i think. Everything is already somewhat following the convention: `mctpi2c%d`, `mctpserial%d`, ...

I could probably open a PR for this if you think this is ok.

jk-ozlabs: Yeah, we'd need the kernel to provide the physical medium identifier. It doesn't belong in the ops struct, but there are certainly ways we can provide that to the `struct mctp_dev` in the kernel (and then on to the netlink interface as a new attribute; say `IFLA_MCTP_PHYS_MEDIUM`).

Given that:

1. the PCIe VDM transport is the only one that requires a Discovery Notify
2. that transport driver is not yet upstream
3. none of the currently-upstream transport drivers provide bindings that require Discovery Notify messages

\- then I would suggest that we add the infrastructure for `IFLA_MCTP_PHYS_MEDIUM`, and have your driver provide the attribute. Then, `mctpd` will only send Discovery Notify messages attribute is present (ie., we know the medium type) and matches the media types that need a Discovery Notify.

Since the Discovery Notify is only required for the non-upstream transport, you'll need kernel patches *anyway*.

> we may temporarily provide such information in mctpd by basing on the interface name i think. Everything is already somewhat following the convention: mctpi2c%d, mctpserial%d, ...

Interface names are not a stable API that we can rely on; they can be renamed arbitrarily.

I'm happy to do the IFLA_MCTP_PHYS_MEDIUM support; drivers will just need a minor change to populate the type value.

src/mctpd.c
```diff
@@ -2895,6 +2895,48 @@ static int bus_link_get_prop(sd_bus *bus,
...
+	dest_phys desti = { 0 }, *dest = &desti;
+	struct mctp_ctrl_cmd_discovery_notify req = { 0 };
+	struct mctp_ctrl_resp_discovery_notify *resp;
+	uint8_t *buf;
+	size_t buf_size;
+	struct sockaddr_mctp_ext resp_addr;
+
+	dest->ifindex = ifindex;
+	mctp_nl_ifaddr_byindex(ctx->nl, dest->ifindex, &dest->hwaddr_len);
+	memset(dest->hwaddr, 0, sizeof dest->hwaddr);
```
jk-ozlabs, line 2910: A zero hwaddr is specific to PCIe, right? Do we also need to know the physical broadcast address for this link type?

src/mctpd.c
khangng-ampere: > I'm happy to do the IFLA_MCTP_PHYS_MEDIUM support; drivers will just need a minor change to populate the type value.

That would be great 😀 Thank you!

tests/test_mctpd_endpoint.py
```diff
@@ -113,12 +118,37 @@ class TestDiscovery:
     async def iface(self):
         return System.Interface("mctp0", 1, 1, bytes([0x1D]), 68, 254, True, PhysicalBinding.PCIE_VDM)
 
+    @pytest.fixture
+    async def bo(self, iface):
+        return TestDiscovery.BusOwnerEndpoint(iface, bytes([0x00]), eid=8)
```
khangng-ampere, line 123: I think I might need to add support for other routing type here like broadcast, or specific route-to-root-complex type inside `Network`. Any ideas here? (Currently I'm just using zeroes for the bus owner physical address 😄 )

khangng-ampere (commented):
Refreshing this 2 years old patch 😀

src/mctpd.c
```diff
@@ -4309,7 +4391,29 @@ static int add_interface(struct ctx *ctx, int ifindex)
...
-		link->discovered = DISCOVERY_UNDISCOVERED;
+		link->discovery.flag = DISCOVERY_UNDISCOVERED;
+		// TODO: These numbers are respectively MN1 and MT4, specified in DSP0239
+		// control message timing.
+		//
+		// Might need to extract these to macros like MCTP_I2C_TSYM_* in this file,
+		// or a commit to actually centralize those timing at one place, now that
+		// we have support for detecting link binding type.
+		link->discovery.notify_retries_left = 2;
+		link->discovery.notify_retry_delay = 5000000;
```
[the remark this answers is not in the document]
jk-ozlabs: Yep, I'd like to implement time control through the ops interface; it'd just be a `get_time()` (or similar) callback; mctpd would then use that to set the poll timeouts (as it does currently, just with no callback).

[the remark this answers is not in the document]
jk-ozlabs: no wait, that might be a bit more complext with the `sd_event` bits, let me check it out in more detail.

tests/test_mctpd_endpoint.py
khangng-ampere: Yeah, extending that in `Endpoint` does look really nice!

jk-ozlabs: Yep, that's fine.

src/mctpd.c
```diff
@@ -4526,7 +4615,42 @@ static int add_interface(struct ctx *ctx, int ifindex)
 	}
 
 	if (phys_binding == MCTP_PHYS_BINDING_PCIE_VDM) {
-		link->discovered = DISCOVERY_UNDISCOVERED;
+		link->discovery.flag = DISCOVERY_UNDISCOVERED;
```
jk-ozlabs: We want the discovery process (below) to work for any binding type that supports Discovery Notify, not just VDM, right?

Should we move that out to be conditional on `link->discovery.flag == DISCOVERY_UNDISCOVERED` ?

(there may be some binding-specific addressing info needed, but the rest should be common?)

khangng-ampere:
I think I will look into replacing PCIe-VDM with I3C in this PR. I3C supports Discovery Notify, already upstreamed, so the address should be standard. The process begins with a Get MCTP Version though, so I will need to tweak it a bit.

PR #16, opened 2023-10-06
mctpd: Handle discovery messages

khangng-ampere:
This MR adds support for discovery messages, when `mctpd` runs on an endpoint.

### DIFF
--- a/src/mctp-netlink.c
+++ b/src/mctp-netlink.c
@@ -1056,6 +1056,76 @@ static int linkmap_add_entry(mctp_nl *nl, struct ifinfomsg *info,
 	return 0;
 }
 
+/* Common parts of RTM_NEWADDR and RTM_DELADDR */
+struct mctp_addralter_msg {
+	struct nlmsghdr nh;
+	struct ifaddrmsg ifmsg;
+	struct rtattr rta;
+	uint8_t data[4];
+};
+static int fill_addralter_args(struct mctp_nl *nl,
+			       struct mctp_addralter_msg *msg,
+			       struct rtattr **prta, size_t *prta_len,
+			       mctp_eid_t eid, const char *linkstr)
+{
+	int ifindex;
+
+	ifindex = mctp_nl_ifindex_byname(nl, linkstr);
+	if (!ifindex) {
+		warnx("invalid device %s", linkstr);
+		return -1;
+	}
+
+	memset(msg, 0x0, sizeof(*msg));
+
+	msg->nh.nlmsg_flags = NLM_F_REQUEST | NLM_F_ACK;
+
+	msg->ifmsg.ifa_index = ifindex;
+	msg->ifmsg.ifa_family = AF_MCTP;
+
+	msg->rta.rta_type = IFA_LOCAL;
+	msg->rta.rta_len = RTA_LENGTH(sizeof(eid));
+	memcpy(RTA_DATA(&msg->rta), &eid, sizeof(eid));
+
+	msg->nh.nlmsg_len =
+		NLMSG_LENGTH(sizeof(msg->ifmsg)) + RTA_SPACE(sizeof(eid));
+
+	if (prta)
+		*prta = &msg->rta;
+	if (prta_len)
+		*prta_len = msg->rta.rta_len;
+
+	return 0;
+}
+
+int mctp_nl_addr_add(struct mctp_nl *nl, mctp_eid_t eid, const char *linkstr)
+{
+	struct mctp_addralter_msg msg;
+	int rc;
+
+	rc = fill_addralter_args(nl, &msg, NULL, NULL, eid, linkstr);
+	if (rc)
+		return -1;
+
+	msg.nh.nlmsg_type = RTM_NEWADDR;
+
+	return mctp_nl_send(nl, &msg.nh);
+}
+
+int mctp_nl_addr_del(struct mctp_nl *nl, mctp_eid_t eid, const char *linkstr)
+{
+	struct mctp_addralter_msg msg;
+	int rc;
+
+	rc = fill_addralter_args(nl, &msg, NULL, NULL, eid, linkstr);
+	if (rc)
+		return -1;
+
+	msg.nh.nlmsg_type = RTM_DELADDR;
+
+	return mctp_nl_send(nl, &msg.nh);
+}
+
 /* Common parts of RTM_NEWROUTE and RTM_DELROUTE */
 struct mctp_rtalter_msg {
 	struct nlmsghdr		nh;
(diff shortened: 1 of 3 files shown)


jk-ozlabs:
> @jk-ozlabs Sorry for the late reply. I made a MR to add the BO configuration at #14, if that is what you meant.

Ah! Thank you. I should have started there.

jk-ozlabs:
How does this relate to #16? should I check that one out first, or this one?

khangng-ampere:
> How does this relate to #16? should I check that one out first, or this one?

@jk-ozlabs Do you mean #96 (set endpoint ID)? This one is based on top of Set Endpoint ID so you should review that one first.

Discovery process contains setting EID, so I think basing this on Set Endpoint ID makes more sense. Set Endpoint ID also works standalone, and not every transport supports Set Endpoint ID

jk-ozlabs:
The DSP0236 implies that we should only support discovery on certain link types:

> Note that Discovered flag is only used for some physical transport
> bindings. An ERROR_INVALID_DATA completion code shall be returned
> if this operation is selected and the particular transport binding does not
> support a Discovered flag.

any thoughts on making this selective?

khangng-ampere:
> The DSP0236 implies that we should only support discovery on certain link types:
> 
> > Note that Discovered flag is only used for some physical transport
> > bindings. An ERROR_INVALID_DATA completion code shall be returned
> > if this operation is selected and the particular transport binding does not
> > support a Discovered flag.
> 
> any thoughts on making this selective?

I think one option is to make `link->discovered` tri-state: unsupported, undiscovered and discovered. That helps decouple the state initialization to when we add the link, where we can base on the link medium type to decide if it is supported or not. Other parts of mctpd can check if this flag is supported. What do you think?

khangng-ampere:
I suppose it would look something like this. There is still no link medium check, but at least bus owner mode should not be impacted

jk-ozlabs:
I think that looks reasonable; we would just set up the unsupported/undiscovered state depending on link type.

My main concern is not breaking things when we (eventually) use the medium type to detect this as some transports will no longer support discovery at that point. A couple of options:

 - we just do the transport detection now
 - we add this, but all links come up as `DISCOVERY_UNSUPPORTED` for now (but that may make testing somewhat difficult)
 - we accept the future breakage
 - we only enable this with a config option
 - something else?

khangng-ampere:
> I think that looks reasonable; we would just set up the unsupported/undiscovered state depending on link type.
> 
> My main concern is not breaking things when we (eventually) use the medium type to detect this as some transports will no longer support discovery at that point. A couple of options:
> 
> * we just do the transport detection now
> * we add this, but all links come up as `DISCOVERY_UNSUPPORTED` for now (but that may make testing somewhat difficult)
> * we accept the future breakage
> * we only enable this with a config option
> * something else?

I think I should just support the transport detection now, since I will have to do that sooner or later anyway. Let me update the PR

src/mctp-netlink.c
```diff
@@ -785,11 +789,14 @@ static int parse_getlink_dump(mctp_nl *nl, struct nlmsghdr *nlh, uint32_t len)
...
-		/* TODO: media type */
+		/* Treat missing physical binding type attribute as UNSPEC */
+		phys_binding = MCTP_PHYS_BINDING_UNSPEC;
+		mctp_get_rtnlmsg_attr_u8(IFLA_MCTP_PHYS_BINDING, rt_mctp, mlen,
+					 &phys_binding);
 
 		up = info->ifi_flags & IFF_UP;
 		linkmap_add_entry(nl, info, ifname, ifname_len, net, up,
-				  min_mtu, max_mtu, hwaddr_len);
+				  min_mtu, max_mtu, hwaddr_len, phys_binding);
```
jk-ozlabs, line 799: This is an arbitrarily-valued `uint8_t`, which you're now casting to an `enum mctp_phys_binding`. With the latter, we would expect it to be one of the valid binding types, which this may not be.

Should we just store this as a uint8_t instead maybe? Or we could convert unknown to `UNSPEC`...

tests/mctpenv/__init__.py
```diff
@@ -37,13 +38,26 @@ def to_nlmsg(self, seq = 0):
...
+    VENDOR = 0xFF
+
 class System:
     class Interface:
         """Interface constructor.
 
         Initial mtu is set to max_mtu.
         """
-        def __init__(self, name, ifindex, net, lladdr, min_mtu, max_mtu, up = False):
+        def __init__(self, name, ifindex, net, lladdr, min_mtu, max_mtu, up = False, phys_binding = PhysicalBinding.UNSPEC):
```
jk-ozlabs, line 60: please wrap to 80 cols. I'll add some proper python formatting at some point, but best to keep with rough conventions in the meantime.

src/mctpd.c
```diff
@@ -3937,6 +4064,11 @@ static int add_interface(struct ctx *ctx, int ifindex)
 					 bus_link_owner_vtable, link);
 	}
 
+	if (phys_binding == MCTP_PHYS_BINDING_PCIE_VDM ||
+	    phys_binding == MCTP_PHYS_BINDING_I3C) {
```
jk-ozlabs, line 4068: From DSP0233 (MCTP over I3C transport binding):

> Note that Prepare for Endpoint Discovery or Endpoint Discovery MCTP control commands are not used to discover MCTP endpoints.

so maybe we don't want this here? We'd need some other mechanism for Discovery Notify (which *is* used for i3c) perhaps?

src/mctp-netlink.c
khangng-ampere: That is true, I think let's just store `uint8_t` and consider the enum values as known constants, not exhaustive

src/mctpd.c
khangng-ampere: Ah dang, I thought it would require the full set of discovery commands 😀 You're right, only Discovery Notify is utilized in I3C

tests/test_mctpd_endpoint.py
```diff
@@ -18,7 +18,7 @@ def config():
 @pytest.fixture
 async def sysnet():
     system = System()
-    iface = System.Interface("mctp0", 1, 1, bytes([0x1D]), 68, 254, True)
+    iface = System.Interface("mctp0", 1, 1, bytes([0x1D]), 68, 254, True, PhysicalBinding.PCIE_VDM)
```
jk-ozlabs, line 21: I'm a little wary of making this the default, as there is no upstream support for the PCIe binding type - we want to make sure that if there are any cases we might hit with common scenarios, that we are likely to cover those in tests.

Is it possible to set this in the test cases where required?

khangng-ampere: That is reasonable yeah. I scoped those test in classes and used `pytest` fixtures to override the interface binding types

jk-ozlabs: nice!

PR #17, opened 2023-10-17
Support get static eid config from entity manager

BonnieLo:
The mctpd will try to find the configuration interface "xyz.openbmc_project.Configuration.MctpDevice" and parse the data. Then it will send get EID command to target device. If we the get the response, we publish the EID on the D-Bus.

Test log:
- reboot and check MCTP D-Bus path root@bmc:~# reboot

root@bmc:~# busctl tree xyz.openbmc_project.MCTP
`- /xyz
  `- /xyz/openbmc_project `- /xyz/openbmc_project/mctp `- /xyz/openbmc_project/mctp/1 |- /xyz/openbmc_project/mctp/1/60 `- /xyz/openbmc_project/mctp/1/8 root@bmc:~# pldmtool base GetTID -m 60
{
    "Response": 134
}

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -38,6 +38,12 @@
 #define MCTP_DBUS_IFACE "xyz.openbmc_project.MCTP"
 #define MCTP_DBUS_IFACE_ENDPOINT "xyz.openbmc_project.MCTP.Endpoint"
 #define OPENBMC_IFACE_COMMON_UUID "xyz.openbmc_project.Common.UUID"
+#define OPENBMC_MCTP_CONFIG_IFACE "xyz.openbmc_project.Configuration.MCTPEndpoint"
+#define OPENBMC_ASSOCIATION_IFACE "xyz.openbmc_project.Association.Definitions"
+#define OPENBMC_ENTITY_MANAGER "xyz.openbmc_project.EntityManager"
+#define OPENBMC_OBJECT_MAPPER "xyz.openbmc_project.ObjectMapper"
+#define OPENBMC_OBJECT_MAPPER_OBJECT_PATH "/xyz/openbmc_project/object_mapper"
+#define OPENBMC_OBJECT_MAPPER_IFACE "xyz.openbmc_project.ObjectMapper"
 
 // an arbitrary constant for use with sd_id128_get_machine_app_specific()
 static const char* mctpd_appid = "67369c05-4b97-4b7e-be72-65cfd8639f10";
@@ -47,6 +53,7 @@ static mctp_eid_t eid_alloc_max = 0xfe;
 
 // arbitrary sanity
 static size_t MAX_PEER_SIZE = 1000000;
+static size_t MAX_STRING_SIZE = 256;
 
 static const uint8_t RQDI_REQ = 1<<7;
 static const uint8_t RQDI_RESP = 0x0;
@@ -70,6 +77,13 @@ struct ctx;
 
 // all local peers have the same phys
 static const dest_phys local_phys = { .ifindex = 0 };
+
+struct association {
+	char *forward;
+	char *reverse;
+	char *object_path;
+};
+typedef struct association association;
 
 struct peer {
 	int net;
@@ -109,6 +123,9 @@ struct peer {
 
 	// From Get Endpoint UUID. A malloced 16 bytes */
 	uint8_t *uuid;
+
+	// the association is an struct of (forward, reverse, object_path)
+	association *association;
 
 	// Stuff the ctx pointer into peer for tidier parameter passing
 	struct ctx *ctx;
@@ -211,6 +228,18 @@ static peer * find_peer_by_addr(ctx *ctx, mctp_eid_t eid, int net)
 
 	if (eid != 0 && n && n->peeridx[eid] >= 0)
 		return &ctx->peers[n->peeridx[eid]];
+	return NULL;
+}
+
+static peer * find_peer_by_association(ctx *ctx, const char *object_path)
+{
+	for (size_t i = 0; i < ctx->size_peers; i++) {
+		peer *peer = &ctx->peers[i];
+		if (peer->state != REMOTE)
+			continue;
+		if (strcmp(peer->association->object_path, object_path) == 0)
+			return peer;
+	}
 	return NULL;
 }
 
@@ -1121,7 +1150,7 @@ out:
 /* Returns the newly added peer.
  * Error is -EEXISTS if it exists */
 static int add_peer(ctx *ctx, const dest_phys *dest, mctp_eid_t eid,
-	int net, peer **ret_peer)
+	int net, char *object_path, peer **ret_peer)
 {
 	ssize_t idx;
 	size_t new_size;
@@ -1183,6 +1212,14 @@ static int add_peer(ctx *ctx, const dest_phys *dest, mctp_eid_t eid,
 	// Update network eid map
 	n->peeridx[eid] = idx;
 
+	// Set association object path
+	if (object_path) {
+		peer->association = malloc(sizeof(struct association));
+		peer->association->forward = strdup("chassis");
+		peer->association->reverse = strdup("mctp_endpoints");
+		peer->association->object_path = strdup(object_path);
+	}
+
 	*ret_peer = peer;
 	return 0;
 }
@@ -1243,6 +1280,12 @@ static int remove_peer(peer *peer)
 	n->peeridx[peer->eid] = -1;
 	free(peer->message_types);
 	free(peer->uuid);
+	if (peer->association) {
+		free(peer->association->forward);
+		free(peer->association->reverse);
+		free(peer->association->object_path);
+		free(peer->association);
+	}
 	memset(peer, 0x0, sizeof(struct peer));
 	return 0;
 }
@@ -1324,7 +1367,7 @@ static int endpoint_assign_eid(ctx *ctx, sd_bus_error *berr, const dest_phys *de
 	/* Find an unused EID */
 	for (e = eid_alloc_min; e <= eid_alloc_max; e++) {
 		if (n->peeridx[e] == -1) {
-			rc = add_peer(ctx, dest, e, net, &peer);
+			rc = add_peer(ctx, dest, e, net, NULL, &peer);
 			if (rc < 0)
 				return rc;
 			break;
@@ -1503,7 +1546,7 @@ static int get_endpoint_peer(ctx *ctx, sd_bus_error *berr,
 			return 0;
 		}
 		/* New endpoint */
-		rc = add_peer(ctx, dest, eid, net, &peer);
+		rc = add_peer(ctx, dest, eid, net, NULL, &peer);
 		if (rc == -EEXIST)
 			return sd_bus_error_setf(berr, SD_BUS_ERROR_FILE_EXISTS,
 					"Endpoint claimed EID %d which is already used",
@@ -1664,6 +1707,160 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
 	memcpy(dest->hwaddr, msg_hwaddr, msg_hwaddr_len);
 	dest->hwaddr_len = msg_hwaddr_len;
 	return 0;
+}
+
+static int setup_static_eid(ctx *ctx)
+{
+	char **object_paths;
+	int net;
+	dest_phys desti = {0}, *dest = &desti;
+	peer *peer = NULL;
+	struct sockaddr_mctp_ext addr;
+	struct mctp_ctrl_cmd_get_eid req = {0};
+	struct mctp_ctrl_resp_get_eid *resp = NULL;
+	req.ctrl_hdr.rq_dgram_inst = RQDI_REQ;
+	req.ctrl_hdr.command_code = MCTP_CTRL_CMD_GET_ENDPOINT_ID;
+	sd_bus_error error = SD_BUS_ERROR_NULL;
+	sd_bus_message *message = NULL;
+
+	int rc = sd_bus_call_method(ctx->bus,
+		OPENBMC_OBJECT_MAPPER, OPENBMC_OBJECT_MAPPER_OBJECT_PATH,
+		OPENBMC_OBJECT_MAPPER_IFACE, "GetSubTreePaths",
+		&error, &message, "sias", "/xyz/openbmc_project/", 0, 1,
+		OPENBMC_MCTP_CONFIG_IFACE);
+	if (rc < 0) {
+		warnx("Failed to issue method call: %s", error.message);
+		goto out;
+	}
+
+	sd_bus_message_read_strv(message, &object_paths);
+	message = sd_bus_message_unref(message);
+
+	// Check if there is configuration found
+	if (!object_paths)
+		goto out;
+
+	for (int i = 0; object_paths[i]; i++) {
+		rc = sd_bus_get_property(ctx->bus,
+			OPENBMC_ENTITY_MANAGER, object_paths[i],
+			OPENBMC_MCTP_CONFIG_IFACE, "EndpointId",
+			&error, &message, "t");
+		if (rc < 0) {
+			warnx("Failed to get endpoint ID from %s. Error:%s", object_paths[i],
+				error.message);
+			break;
+		}
+		uint64_t data = 0;
+		sd_bus_message_read(message, "t", &data);
+		mctp_eid_t eid = (mctp_eid_t)data;
+
+		peer = find_peer_by_addr(ctx, eid, net);
+		if (peer) {
+			// Endpoint already added
+			continue;
+		}
+
+		message = sd_bus_message_unref(message);
+
+		rc = sd_bus_get_property(ctx->bus,
+			OPENBMC_ENTITY_MANAGER, object_paths[i],
+			OPENBMC_MCTP_CONFIG_IFACE, "Address",
+			&error, &message, "t");
+		if (rc < 0) {
+			warnx("Failed to get address from %s. Error:%s", object_paths[i],
+				error.message);
+			break;
+		}
+
+		sd_bus_message_read(message, "t", &data);
+
+		size_t size = sizeof(uint8_t);
+		memcpy(dest->hwaddr, &data, size);
+		dest->hwaddr_len = size;
+
+		message = sd_bus_message_unref(message);
+
+		rc = sd_bus_get_property(ctx->bus,
+			OPENBMC_ENTITY_MANAGER, object_paths[i],
+			OPENBMC_MCTP_CONFIG_IFACE, "Bus",
+			&error, &message, "t");
+		if (rc < 0) {
+			warnx("Failed to get bus from %s. Error:%s", object_paths[i],
+				error.message);
+			break;
+		}
+		uint64_t bus = 0;
+		char ifname[MAX_STRING_SIZE];
+		sd_bus_message_read(message, "t", &bus);
+		// TODO: handle different interface name
+		sprintf(ifname, "mctpi2c%ld", bus);
+
+		dest->ifindex = mctp_nl_ifindex_byname(ctx->nl, ifname);
+		if (dest->ifindex <= 0) {
+			warnx("Unknown ifname: %s", ifname);
+			continue;
+		}
+
+		rc = validate_dest_phys(ctx, dest);
+		if (rc < 0) {
+			warnx("Bad phys: 0x%02x", dest->hwaddr[0]);
+			continue;
+		}
+
+		net = mctp_nl_net_byindex(ctx->nl, dest->ifindex);
+		if (net < 1) {
+			warnx("No net for ifindex");
+			continue;
+		}
+
+		message = sd_bus_message_unref(message);
+
+		// The s[i] is like ../chassis/device so we need to get its parent path
+		// for the chassis path.
+		char object_path[MAX_STRING_SIZE];
+		if (strlen(object_paths[i]) >= MAX_STRING_SIZE) {
+			warnx("Object path too large >= %d", MAX_STRING_SIZE);
+			continue;
+		}
+		strcpy(object_path, object_paths[i]);
+		char *ret = strrchr(object_path, '/');
+		if (ret)
+			*ret = '\0';
+
+		rc = add_peer(ctx, dest, eid, net, object_path, &peer);
+		if (rc < 0) {
+			warnx("Failed to add peer for EID %d", eid);
+			continue;
+		}
+
+		add_peer_route(peer);
+
+		uint8_t* buf = NULL;
+		size_t buf_size = 0;
+		rc = endpoint_query_peer(peer, MCTP_CTRL_HDR_MSG_TYPE,
+			&req, sizeof(req), &buf, &buf_size, &addr);
+		if (rc < 0) {
+			warnx("Response timeout for EID %d.", eid);
+			remove_peer(peer);
+			free(buf);
+			continue;
+		}
+		free(buf);
+
+		rc = query_peer_properties(peer);
+
+		rc = publish_peer(peer, true);
+		if (rc < 0) {
+			warnx("Error publishing remote eid %d net %d", eid, net);
+			continue;
+		}
+
+		message = sd_bus_message_unref(message);
+	}
+
+out:
+	sd_bus_error_free(&error);
+	return rc;
 }
 
 /* SetupEndpoint method tries the following in order:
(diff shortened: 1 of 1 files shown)


williamspatrick:
@jk-ozlabs @mkj - Is there any feedback on this PR?  It has been open for a month with no action.

jk-ozlabs:
> @jk-ozlabs @mkj - Is there any feedback on this PR? It has been open for a month with no action.

So there was a discussion in OpenBMC gerrit, at https://gerrit.openbmc.org/c/openbmc/entity-manager/+/66750, you mentioned:

> I really don't like that we have undocumented dbus interfaces now spanning even outside the OpenBMC organization. 
> Should we start documenting these EM config types in phosphor-dbus-interfaces? (@edtanous - opinions?)

And I replied:

> Yep, these should be tracked with the rest of the config structure.
> 
> Alternatively, we could make a policy that the EM config types don't leak out to other projects (or at least those we have input on), and provide a more "generic" configuration mechanism for mctpd.

So: this is pending a decision on whether we want to expose the EM interfaces outside of OpenBMC.

If not, we'll need a bit of bridge code, within OpenBMC, to perform configuration on non-openbmc code (this change is essentially a shortcut for calling the SetupEndpoint method based on EM config, which could justifiable be done elsewhere)

If so, we should be fine with this as-is.

Let me know what your preference with the interface.

PeterHo-wiwynn:
Hi @jk-ozlabs,
Since the patch https://gerrit.openbmc.org/c/openbmc/entity-manager/+/66750 has been merged, would you mind take a look at this PR? I did added some changes to this PR, so it is a little different from the origin one. The commit message lists all the changes in this PR.

jk-ozlabs:
> Hi @jk-ozlabs, Since the patch https://gerrit.openbmc.org/c/openbmc/entity-manager/+/66750 has been merged, would you mind take a look at this PR?

Sure thing! I'll get that done this week.

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+
+	int i = 0;
+	size_t size = 0;
+	dest_phys desti = {0}, *dest = &desti;
+	peer *peer = NULL;
+	int net;
+	struct sockaddr_mctp_ext addr;
+	struct mctp_ctrl_cmd_get_eid req = {0};
+	struct mctp_ctrl_resp_get_eid *resp = NULL;
+	char *peer_path = NULL;
```
jk-ozlabs, line 1732: Fairly minor, but can you keep the declarations to the top of the defining block?

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
 
+static int setup_static_eid(ctx *ctx)
+{
+	int rc;
+	sd_bus_error error = SD_BUS_ERROR_NULL;
+	sd_bus_message *message = NULL;
+	char **s;
+
+	rc = sd_bus_call_method(ctx->bus,
+        "xyz.openbmc_project.ObjectMapper", "/xyz/openbmc_project/object_mapper",
```
jk-ozlabs, line 1708: Minor: Indentation is a little off here.

Also, might be worthwhile adding `const`s or `#define`s for these bus names / object paths / interfaces.

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+			OPENBMC_MCTP_CONFIG_IFACE, "Bus",
+			&error, &message, "t");
+		if (rc < 0) {
+			warnx("Failed to get property: %s", error.message);
+			goto out;
+		}
+		uint64_t bus = 0;
+		char *ifname = NULL;
+		sd_bus_message_read(message, "t", &bus);
+		ifname = malloc(11); // length for "mctpi2cxxx"
```
jk-ozlabs, line 1768: Just use the stack for this.

Also, you'll need more than 11 chars for all possible values of that `uint64_t`

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+		snprintf(ifname, 11, "mctpi2c%d", (uint8_t)bus);
```
jk-ozlabs, line 1769: We might want a todo here, to address later. This isn't guaranteed to be the interface name.

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+		if (rc < 0) {
+			warnx("Failed to get property: %s\n", error.message);
+			goto out;
+		}
+
+		const uint64_t data = 0;
+		sd_bus_message_read(message, "t", &data);
+
+		size = sizeof(uint8_t);
+		memset(dest->hwaddr, 0x0, MAX_ADDR_LEN);
```
jk-ozlabs, line 1751: This will already be zeroed above.

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+		if (rc < 0) {
+			warnx("Failed to get property: %s\n", error.message);
+			goto out;
+		}
+		uint64_t eid = 0;
+		sd_bus_message_read(message, "t", &eid);
+
+		rc = add_peer(ctx, dest, (mctp_eid_t)eid, net, &peer);
+		if (rc < 0) {
+			warnx("Failed to add peer for EID %d", (mctp_eid_t)eid);
```
jk-ozlabs, line 1808: maybe do the cast once after extracting from the `sb_bus_message_read()`, instead of every subsequent use of `eid`.

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+		rc = query_peer_properties(peer);
+
+		// The s[i] is like ../chassis/device so we need to get its parent path
+		// for the chassis path.
+		char object_path[100];
+		strcpy(object_path, s[i]);
+		char *ret = strrchr(object_path, '/');
+		*ret = '\0';
+
+		peer->association = malloc(sizeof(struct association));
```
jk-ozlabs, line 1837: We could have this allocated as part of the `peer`, and then just use a boolean flag to indicate no association.

src/mctpd.c
```diff
@@ -1832,6 +2021,18 @@ static int method_learn_endpoint(sd_bus_message *call, void *data, sd_bus_error
 	return rc;
 }
 
+static int method_setup_endpoint_by_config(sd_bus_message *call, void *data, sd_bus_error *berr)
+{
+	int rc = 0;
+	ctx *ctx = data;
+	rc = setup_static_eid(ctx);
```
jk-ozlabs, line 2028: no handling here? if not this is what you intended, then drop the assignment.

jk-ozlabs (changes requested):
Hi!

Thanks for the change, overall I'm happy with the strategy here.

There are a few comments on the code though. Most are super minor, except for the InterfacesAdded / InterfacesRemoved handling - if possible, we probably want to filter those at the match level too, so we're not getting floods of signals.

src/mctpd.c
```diff
@@ -2584,6 +2823,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+
+	rc = sd_bus_message_enter_container(m, SD_BUS_TYPE_ARRAY, "s");
+	if (rc < 0)
+		goto out;
+
+	for(;;) {
+		rc = sd_bus_message_read(m, "s", &iface);
+		if (rc == 0)
+			break;
+		if(strcmp(iface,OPENBMC_MCTP_CONFIG_IFACE) == 0) {
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -2584,6 +2823,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+			update = true;
+		}
+	}
+
+	if (update == true) {
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -2584,6 +2823,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+		char temp[100];
+		strcpy(temp, object_path);
+		char *ret = strrchr(temp, '/');
+		*ret = '\0';
+
+		for(;;) {
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -2668,6 +3004,36 @@ static int setup_bus(ctx *ctx)
...
+	rc = sd_bus_match_signal(
+		ctx->bus,
+		NULL,
+		NULL,
+		"/xyz/openbmc_project/inventory",
+		"org.freedesktop.DBus.ObjectManager",
+		"InterfacesAdded",
+		on_interface_added,
+		ctx
+	);
```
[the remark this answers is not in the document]
PeterHo-wiwynn: I have tried `arg[0, ...]` and `arg0namespace` but didn't work well. Do you have any idea how to filter it?
I also checked how others OpenBMC's code handle it. They use callback function to filter the signal instead of filtering signals by match.

src/mctpd.c
```diff
@@ -2668,6 +3004,36 @@ static int setup_bus(ctx *ctx)
...
+	rc = sd_bus_match_signal(
+		ctx->bus,
+		NULL,
+		NULL,
+		"/xyz/openbmc_project/inventory",
+		"org.freedesktop.DBus.ObjectManager",
+		"InterfacesRemoved",
+		on_interface_removed,
+		ctx
+	);
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Same as above, might need your feedback on this.

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -1666,6 +1697,164 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+		free(buf);
+
+		rc = query_peer_properties(peer);
+
+		// The s[i] is like ../chassis/device so we need to get its parent path
+		// for the chassis path.
+		char object_path[100];
+		strcpy(object_path, s[i]);
+		char *ret = strrchr(object_path, '/');
+		*ret = '\0';
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done. I added this to `add_peer`.

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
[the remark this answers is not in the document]
jk-ozlabs: Ah, I had incorrectly assumed that `arg0` would specify the interface name, but it's actually the object path. That may make it difficult to filter. I'll see if there's a better alternative, but we may just have to rely on filtering in the callback (as you have done already).

src/mctpd.c
jk-ozlabs: Why a cast to a uint8_t here? It's not infeasible we'll have a interface name above 255.

src/mctpd.c
```diff
@@ -1666,6 +1708,161 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+			i++;
+			continue;
+		}
+
+		message = sd_bus_message_unref(message);
+
+		// The s[i] is like ../chassis/device so we need to get its parent path
+		// for the chassis path.
+		char object_path[100];
+		strcpy(object_path, object_paths[i]);
```
jk-ozlabs, line 1821: where is `100` from here? looks like this could this cause an overflow?

src/mctpd.c
```diff
@@ -1666,6 +1708,161 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+			i++;
+			continue;
+		}
+		free(buf);
+
+		rc = query_peer_properties(peer);
+
+		rc = publish_peer(peer, true);
+		if (rc < 0) {
+			warnx("Error publishing remote eid %d net %d", (mctp_eid_t)eid, net);
```
jk-ozlabs, line 1852: `eid` is already a `mctp_eid_t`, no need for this cast

src/mctpd.c
```diff
@@ -1666,6 +1708,161 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+		}
+
+		message = sd_bus_message_unref(message);
+
+		rc = sd_bus_get_property(ctx->bus,
+			OPENBMC_ENTITY_MANAGER, object_paths[i],
+			OPENBMC_MCTP_CONFIG_IFACE, "Address",
+			&error, &message, "t");
+		if (rc < 0) {
+			warnx("Failed to get property: %s\n", error.message);
```
jk-ozlabs, line 1769: no `\n`

src/mctpd.c
```diff
@@ -1666,6 +1708,161 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+
+		message = sd_bus_message_unref(message);
+
+		rc = sd_bus_get_property(ctx->bus,
+			OPENBMC_ENTITY_MANAGER, object_paths[i],
+			OPENBMC_MCTP_CONFIG_IFACE, "Bus",
+			&error, &message, "t");
+		if (rc < 0) {
+			warnx("Failed to get property: %s", error.message);
+			goto out;
```
jk-ozlabs, line 1787: `break` ?

src/mctpd.c
```diff
@@ -2337,6 +2553,9 @@ static int bus_endpoint_get_prop(sd_bus *bus,
 	} else if (strcmp(property, "UUID") == 0 && peer->uuid) {
 		const char *s = dfree(bytes_to_uuid(peer->uuid));
 		rc = sd_bus_message_append(reply, "s", s);
+	} else if (strcmp(property, "Associations") == 0 && peer->association) {
+		rc = sd_bus_message_append(reply, "a(sss)",
+			1, peer->association->forward, peer->association->reverse, peer->association->object_path);
```
jk-ozlabs, line 2558: super minor, but can you wrap at 80 here?

src/mctpd.c
```diff
@@ -2425,6 +2654,26 @@ static int bus_endpoint_find_uuid(sd_bus *bus, const char *path,
...
+	const char *interface, void *userdata, void **ret_found,
+	sd_bus_error *ret_error)
+{
+	ctx *ctx = userdata;
+	peer *peer = NULL;
+	int rc;
+
+	rc = peer_from_path(ctx, path, &peer);
+	if (rc >= 0 && peer->published) {
+		if (peer->association) {
```
jk-ozlabs, line 2669: just include this in the conditional above maybe?

src/mctpd.c
```diff
@@ -2584,6 +2833,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+			sd_bus_message_skip(m, "a{sv}");
+		}
+		else {
+			sd_bus_message_skip(m, "a{sv}");
+			update = true;
+		}
+		sd_bus_message_exit_container(m);
+	}
+	sd_bus_message_exit_container(m);
+	//TODO: parse message from signal instead of recanning
```
jk-ozlabs, line 2870: What do you mean by 'recanning' here?

src/mctpd.c
```diff
@@ -2584,6 +2833,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+	if (update == true)
+		setup_static_eid(ctx);
+
+out:
```
jk-ozlabs, line 2874: do you really need this label? there's no cleanup code, you could just return directly from the error cases.

src/mctpd.c
jk-ozlabs: You're also generating the same error message as above, so there will be no way to tell which path failed.

jk-ozlabs (changes requested):
I'll take a look at options for filtering InterfacesAdded, but I think we'll just go with what you have here. A few minor changes included as comments, but looks good so far.

src/mctpd.c
```diff
@@ -71,6 +77,13 @@ struct ctx;
 // all local peers have the same phys
 static const dest_phys local_phys = { .ifindex = 0 };
 
+struct association {
+	char *forward;
+	char *reverse;
+	char *object_path;
+};
+typedef struct association association;
```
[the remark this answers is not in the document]
PeterHo-wiwynn: I was following the others struct. Should I drop this typedef still?

src/mctpd.c
```diff
@@ -2584,6 +2833,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+			break;
+
+		rc = sd_bus_message_read(m, "s", &iface);
+		if (rc == 0)
+			break;
+
+		if (strcmp(iface, OPENBMC_MCTP_CONFIG_IFACE) != 0) {
+			sd_bus_message_skip(m, "a{sv}");
+		}
+		else {
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -2584,6 +2833,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+	sd_bus_message_exit_container(m);
+	//TODO: parse message from signal instead of recanning
+	if (update == true)
+		setup_static_eid(ctx);
+
+out:
+	return rc;
+}
+
+int on_interface_removed(sd_bus_message *m, void *userdata, sd_bus_error *ret_error)
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -2584,6 +2833,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+		}
+		else {
+			sd_bus_message_skip(m, "a{sv}");
+			update = true;
+		}
+		sd_bus_message_exit_container(m);
+	}
+	sd_bus_message_exit_container(m);
+	//TODO: parse message from signal instead of recanning
+	if (update == true)
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -2584,6 +2833,92 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
 	return rc;
 }
 
+int on_interface_added(sd_bus_message *m, void *userdata, sd_bus_error *ret_error)
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -1666,6 +1708,161 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+	}
+
+	sd_bus_message_read_strv(message, &object_paths);
+	message = sd_bus_message_unref(message);
+
+	// Check if there is configuration found
+	if (!object_paths)
+		goto out;
+
+	while(object_paths[i] != NULL) {
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
```diff
@@ -1666,6 +1708,161 @@ static int message_read_hwaddr(sd_bus_message *call, dest_phys* dest)
...
+		rc = sd_bus_get_property(ctx->bus,
+			OPENBMC_ENTITY_MANAGER, object_paths[i],
+			OPENBMC_MCTP_CONFIG_IFACE, "EndpointId",
+			&error, &message, "t");
+		if (rc < 0) {
+			warnx("Failed to get property: %s\n", error.message);
```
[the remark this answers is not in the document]
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done. Removed it.

src/mctpd.c
PeterHo-wiwynn: Sorry for typo. It should be 'rescanning'.

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done. I also revised the message.

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: I just roughly defined a max length for it. I increased it to 256 and I think it is enough for an inventory object path.

> Why a cast to a uint8_t here? It's not infeasible we'll have a interface name above 255.

I dropped the casting and changed it to `sprintf`.

src/mctpd.c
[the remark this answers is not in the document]
jk-ozlabs: yeah, you're right, there are plenty of typedefs already. good to keep it consistent.

src/mctpd.c
jk-ozlabs: It may be enough for an inventory object string, but it still may overflow the stack if a message is reply is received with a (possibly intentionally) oversized object path there.

However, we can handle this in a separate change.

src/mctpd.c
```diff
@@ -2425,6 +2650,24 @@ static int bus_endpoint_find_uuid(sd_bus *bus, const char *path,
...
+	const char *interface, void *userdata, void **ret_found,
+	sd_bus_error *ret_error)
+{
+	ctx *ctx = userdata;
+	peer *peer = NULL;
+	int rc;
+
+	rc = peer_from_path(ctx, path, &peer);
+	if (rc >= 0 && peer->published && peer->association) {
+			*ret_found = peer;
```
jk-ozlabs, line 2665: only minor, but indentation is off here.

src/mctpd.c
```diff
@@ -2584,6 +2827,89 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+			update = true;
+		}
+	}
+
+	if (!update) {
+		return rc;
+	}
+
+	char temp[100];
+	strcpy(temp, object_path);
```
jk-ozlabs, line 2900: Same potential overflow here

jk-ozlabs (commented):
Looking good now - just have a couple of potential stack overflows to resolve. If you'd prefer to do those in a separate change, let me know, and I'll merge this and address them separately.

jk-ozlabs:
And one last thing: can you add a `Signed-off-by` line, to indicate that you agree to the DCO: https://developercertificate.org/

ThuBaNguyen:
In which condition the Static Endpoint object path will be created? Because the https://gerrit.openbmc.org/c/openbmc/entity-manager/+/66750 creates the Association without whenever the configuration is existing.

src/mctpd.c
PeterHo-wiwynn: I made it check the length before doing further operation.

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: I also made it check the length here.

PeterHo-wiwynn:
Hi @ThuBaNguyen,

We made it sent the MCTP `Get Endpoint ID` before publishing `EID` on D-Bus. It only publishes the device which has response.

src/mctpd.c
jk-ozlabs: You're off by one here. strcpy will write `strlen(src)+1` bytes.

Maybe just make the conditional:

```c
if (strlen(object_paths[i]) >= MAX_STRING_SIZE) {
```

src/mctpd.c
jk-ozlabs: Same off-by-one as above

jk-ozlabs (commented):
Just a couple of really minor things

jk-ozlabs:
Once the above is addressed, I think we're okay here on an implementation level.

However, what hasn't been addressed is whether the design is acceptable and consistent to the OpenBMC approach. This is effectively leaking OpenBMC configuration definitions outside of the OpenBMC project.

(we do have the precedence of exposing an OpenBMC-namespaced Endpoint interface, but that's a fairly clear choice in not intentionally reinventing an interface that exactly matches the data exposed)

I'm not necessarily against adding the configuration mechanisms too, but want to ensure we're making the right (organisational) choices here. Was there any OpenBMC discussion about doing this? Is there a design document around that we can refer to?

@amboar , @williamspatrick , any thoughts?

If there are solid objections, it may be possible to use this same interface/configuration code in a "connector" project, like the phosphor-networkd approach.

amboar:
I think without a strong argument to support the PR I feel the same about containing OpenBMC-isms in the OpenBMC ecosystem. `mctpd` might see heavy use in OpenBMC, but that doesn't mean it should assume an OpenBMC environment. A suggestion might be to add a build option for the feature, but I don't feel the trade-offs are right for that. Either way it feels like it would be externalising an OpenBMC cost onto the maintenance of mctpd.

My preference is for the "connector" approach. phosphor-networkd is mentioned above as an example of the strategy. An option might be to cut out the middleware and have EM drive the mctpd interfaces directly.

jk-ozlabs:
> A suggestion might be to add a build option for the feature, but I don't feel the trade-offs are right for that.

I'd rather a runtime option if we were to take that approach, so I'm not *too* opposed to it.

(regardless, that would be a separate change to this one; needs more design for the correct configuration mechanism)

> Either way it feels like it would be externalising an OpenBMC cost onto the maintenance of mctpd.

While we do ensure that mctpd is not OpenBMC-specific, there is some pragmatism in that's where more than half of the use-cases are; I'm okay with spending time to make that work well, as long as that's reasonable (specifically - keeping up with changes that may be happening inside the project....)

williamspatrick:
I'm not super excited about the EM config, especially since it is undocumented from a dbus perspective, but I'm not sure that we have a better option in the short-term.  The functionality seems more necessary at this point relative to anyone with time to improve the architectural side of it.

src/mctpd.c
PeterHo-wiwynn: Done

src/mctpd.c
PeterHo-wiwynn: Done

jk-ozlabs:
Just been digging through OpenBMC docs for some guidance on this.

Would merging this put us afoul of https://github.com/openbmc/docs/blob/master/meta-layer-guidelines.md#meta-layers-should-not-point-to-openbmc-specific-repositories-outside-of ?

jk-ozlabs:
> The functionality seems more necessary at this point relative to anyone with time to improve the architectural side of it.

I'm open to putting some time into a connector approach, if that's the best way forward...

amboar:
> > The functionality seems more necessary at this point relative to anyone with time to improve the architectural side of it.
> 
> I'm open to putting some time into a connector approach, if that's the best way forward...

Right, so [based on a discussion I had with Ed in #pmci on the OpenBMC Discord](https://discord.com/channels/775381525260664832/778790638563885086/1197003954882613338) he'd prefer to keep `mctpd` details out of EM as much as I'd prefer to keep EM details out of `mctpd`, which leaves us with the connector daemon. The suggestion is to implement it in (the increasingly misnamed) `dbus-sensors` as it already has a bunch of code to deal with EM's ill-defined D-Bus interfaces.

A bit of a side-quest might be to get `dbus-sensors` renamed to `entity-reactors`.

ThuBaNguyen:
> Hi @ThuBaNguyen,
> 
> We made it sent the MCTP `Get Endpoint ID` before publishing `EID` on D-Bus. It only publishes the device which has response.

For the endpoint such as CPU, the device only available when the host is On. The endpoint can off when the Entity-manager adds the CPU's configuration to D-Bus interface. This can causes the failure of mctpd I `Get Endpoint ID`.
This causes the CPU's EID will never add to mctpd D-Bus interface.
It is the same for the hot plug devices.

PeterHo-wiwynn:
> Right, so [based on a discussion I had with Ed in #pmci on the OpenBMC Discord](https://discord.com/channels/775381525260664832/778790638563885086/1197003954882613338) he'd prefer to keep mctpd details out of EM as much as I'd prefer to keep EM details out of mctpd, which leaves us with the connector daemon. The suggestion is to implement it in (the increasingly misnamed) dbus-sensors as it already has a bunch of code to deal with EM's ill-defined D-Bus interfaces.
>
> A bit of a side-quest might be to get dbus-sensors renamed to entity-reactors.

Hi @amboar, 
I'm not sure what we need to do in `dbus-sensors`. Could you provide more detail about it?

> For the endpoint such as CPU, the device only available when the host is On. The endpoint can off when the Entity-manager adds the CPU's configuration to D-Bus interface. This can causes the failure of mctpd I Get Endpoint ID.
This causes the CPU's EID will never add to mctpd D-Bus interface.
It is the same for the hot plug devices.

Hi @ThuBaNguyen,
For the hot plug devices, the `entity-manager` should be able to handle with the FRU information. We have services which rescanning buses for FRU information in `gpio-monitor`.

But for the CPU, we need to figure out a better way if we need to define a static EID for CPU with this implementation.

amboar:
> > Right, so [based on a discussion I had with Ed in #pmci on the OpenBMC Discord](https://discord.com/channels/775381525260664832/778790638563885086/1197003954882613338) he'd prefer to keep mctpd details out of EM as much as I'd prefer to keep EM details out of mctpd, which leaves us with the connector daemon. The suggestion is to implement it in (the increasingly misnamed) dbus-sensors as it already has a bunch of code to deal with EM's ill-defined D-Bus interfaces.
> > A bit of a side-quest might be to get dbus-sensors renamed to entity-reactors.
> 
> Hi @amboar, I'm not sure what we need to do in `dbus-sensors`. Could you provide more detail about it?

Right now you don't need to do anything, I'm doing some prototyping myself :)

amboar:
Okay, here's a proof-of-concept:

https://github.com/amboar/dbus-sensors/commit/dcf1c1b9df97dfa4d8640e8d255cefaed8d087fd

It doesn't yet handle associations between the EM config and the endpoint objects exposed by `mctpd`. However, it does handle configuring `mctpd` with exposed EM configs, and also re-establishing the setup of endpoints objects that are removed by `mctpd` while their EM configs remain exposed.

amboar:
Okay, I've been chipping away at improvements to the `mctpreactor` proof-of-concept. As of patchset 8 in Gerrit it adds and removes the (proxy) associations as necessary, ticking off the first todo above. It also supplies a systemd service unit which should make it a bit more convenient.

Next thing that needs attention is fixing up the schema in EM.

amboar:
alright, I think the `mctpreactor` bits are in decent enough shape for people to review:

- [schemas: Introduce a schema for MCTP devices](https://gerrit.openbmc.org/c/openbmc/entity-manager/+/69600)
- [Introduce MCTP-related interfaces and implementations](https://gerrit.openbmc.org/c/openbmc/dbus-sensors/+/69110)
- [Add mctpreactor for dynamic configuration of MCTP networks](https://gerrit.openbmc.org/c/openbmc/dbus-sensors/+/69111)

I've done a conversion of the Yosemit4 configs to use the new schemas, but I'm not sure the config as-specified was well formed:

- [configurations: Convert Yosemite4 to MCTPDevice schema](https://gerrit.openbmc.org/c/openbmc/entity-manager/+/69601)

jk-ozlabs:
With the `mctpreactor` changes, it seems like we would no longer need this implementation, as the configuration is handled externally. We would need further work for static EID allocations, but that's now in the EM domain.

So, I'll close this now, but @BonnieLo : let me know if I'm incorrect with the above and we can revisit/reopen.

yikaitsaiww:
Hi @jk-ozlabs,

In our project, we have numerous devices on different I3C buses that use the same I3C address. As a temporary solution, we have been using the `SetupEndpointByConfigPath` implementation introduced in this PR to address this issue.

As I know, `mctpreactor` calls `AssignEndpoint` which cannot support multi devices with the same I3C address. For such specific scenarios, additional handling or configuration might be necessary. Do you have any other suggestions on how we can achieve a similar functionality within `entity-manager` or other recpies?

Thank you for your time and consideration.

jk-ozlabs:
> which cannot support multi devices with the same I3C address

Can you expand on this a little? I assume these are on separate i3c busses, right?

jk-ozlabs:
Ah, you have mentioned that they are separate busses, right.

In which case, `AssignEndpoint` is fine; it is specific to a MCTP interface device.

yikaitsaiww:
Thank you for your prompt response! We will double-check this and create a new issue if we have any further questions.

PR #23, opened 2023-12-21
Health polling

amboar:
Define and implement a process for recovering an endpoint or reclaiming its EID.

I've added `docs/endpoint-recovery.md` which discusses the problems and trade-offs involved.

### DIFF
--- a/docs/endpoint-recovery.md
+++ b/docs/endpoint-recovery.md
@@ -1,0 +1,434 @@
+# Endpoint Recovery
+
+<!--toc:start-->
+- [Endpoint Recovery](#endpoint-recovery)
+  - [References](#references)
+  - [Relevant components](#relevant-components)
+  - [Concerns](#concerns)
+  - [Behaviours](#behaviours)
+  - [Relevant MCTP commands, behaviours, and requirement constraints](#relevant-mctp-commands-behaviours-and-requirement-constraints)
+  - [Requesting endpoint recovery from `mctpd`](#requesting-endpoint-recovery-from-mctpd)
+  - [Endpoint health determinations and consequences](#endpoint-health-determinations-and-consequences)
+  - [D-Bus interface discussion](#d-bus-interface-discussion)
+  - [Proposed Design](#proposed-design)
+    - [`.Recover` method design considerations](#recover-method-design-considerations)
+    - [Application behaviours over an endpoint recovery sequence](#application-behaviours-over-an-endpoint-recovery-sequence)
+    - [`mctpd` behaviours over an endpoint recovery sequence](#mctpd-behaviours-over-an-endpoint-recovery-sequence)
+  - [MCTP device lifecycle and consuming events from `mctpd` D-Bus objects](#mctp-device-lifecycle-and-consuming-events-from-mctpd-d-bus-objects)
+    - [An example order of MCTP device events, consumed by `nvmesensor`](#an-example-order-of-mctp-device-events-consumed-by-nvmesensor)
+<!--toc:end-->
+
+Here we capture some reasoning around a design and implementation of device
+presence detection and EID reclaim.
+
+## References
+
+1. [DSP0236 - Management Component Transport Protocol (MCTP) Base Specification][dmtf-dsp0236]
+2. [DSP0237 - Management Component Transport Protocol (MCTP) SMBus/I2C Transport Binding Specification][dmtf-dsp0237]
+3. [DSP0239 - Management Component Transport Protocol (MCTP) IDs and Codes][dmtf-dsp0239]
+3. [Freedesktop: D-Bus Specification][dbus-spec]
+
+[dmtf-dsp0236]: https://www.dmtf.org/sites/default/files/standards/documents/DSP0236_1.3.1.pdf
+[dmtf-dsp0237]: https://www.dmtf.org/sites/default/files/standards/documents/DSP0237_1.2.0.pdf
+[dmtf-dsp0239]: https://www.dmtf.org/sites/default/files/standards/documents/DSP0239_1.10.0.pdf
+[dbus-spec]: https://dbus.freedesktop.org/doc/dbus-specification.html
+
+## Relevant components
+
+1. Endpoint devices, such as NVMe drives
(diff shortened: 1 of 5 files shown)


src/mctpd.c
```diff
@@ -1587,6 +1612,45 @@ static int peer_set_uuid(peer *peer, const uint8_t uuid[16])
 	return 0;
 }
 
+static int
+query_get_peer_uuid_by_phys(ctx *ctx, const dest_phys *dest, uint8_t uuid[16]) {
```
jk-ozlabs, line 1618: Formatting?

docs/endpoint-recovery.md
```diff
@@ -0,0 +1,434 @@
...
+   ```
+
+2. Issue a call to `GetManagedObjects` `org.freedesktop.DBus.ObjectManager` on
+   `/xyz/openbmc_project/mctp` for initial population of local data-structures
+
+3. Issue
+
+### An example order of MCTP device events, consumed by `nvmesensor`
+
+1. A series of events yields a new FRU EEPROM device on an I2C bus already
```
ThuBaNguyen, line 359: The new FRU EEPROM of the device on I2C bus some time does not mean the device is ready for the mctp communication.
For the card which have the embedded OS, the FRU EEPROM will available before the OS is ready to response for MCTP messages.

docs/endpoint-recovery.md
```diff
@@ -0,0 +1,434 @@
...
+   registered as an MCTP endpoint.
+
+2. The FRU data is decoded and `SetupEndpoint` on the
+   `au.com.CodeConstruct.MCTP` interface of the `/xyz/openbmc_project/mctp`
+   object hosted by `mctpd` is invoked.
+
+3. The device was not previously configured and has no programmed static EID.
+   `mctpd` allocates an EID (9) from its dynamic pool and assigns it to the
+   device.
```
ThuBaNguyen, line 368: This step assumed that the device is ready and can response for the `GetEID`, `SetEID` commands. Without the response from the device for the MCTP commands, the step 4 will never be called, right?

docs/endpoint-recovery.md
jk-ozlabs: It's up to external processes to determine when the endpoint is ready (and call SetupEndpoint). In some cases, the presense of the FRU EEPROM is enough. In others it may not be.

I have some plans to add SMBus ARP support to do the discovery, but that's a bit down the pipeline. That would be a different instance of those "external processes".

ThuBaNguyen: `It's up to external processes to determine when the endpoint is ready (and call SetupEndpoint). In some cases, the presense of the FRU EEPROM is enough. In others it may not be.` That means we will have some device which is not ready for MCTP message before step 3. The failure in steps 3 will cause the added EID will removed from the D-Bus and the EID will never be added again.

Because in this case step 1 and 2 are from out side service, maybe we can add one note to request the service make sure that the device is ready for MCTP interface before call `setupEndpoint`.

docs/endpoint-recovery.md
```diff
@@ -0,0 +1,434 @@
...
+   `/xyz/openbmc_project/mctp` for initial population of local data-structures
+
+3. Issue
+
+### An example order of MCTP device events, consumed by `nvmesensor`
+
+1. A series of events yields a new FRU EEPROM device on an I2C bus already
+   registered as an MCTP endpoint.
+
+2. The FRU data is decoded and `SetupEndpoint` on the
```
ThuBaNguyen, line 362: This is an example for the setting which BMC is BO or TMBO. How about in the PCIe interface when BMC can be endpoint?
In that case, instead of detect the FRU, BMC will receive the `prepare for discovery` signal from PCIe BO. As the previous discussion in `pcmi` discord, `mctpd` will response for that command and add the PCIe EID of BMC and also BO to mctp D-Bus interface.
Then `mctpd` will get the routing table from PCIe BO. Then how `mctpd` manages the EID in the received entry in the routing table? Does `mctpd` check the available status of device before adding it the MCTP D-Bus interface?

docs/endpoint-recovery.md
```diff
@@ -0,0 +1,434 @@
...
+
+## Proposed Design
+
+The approach is to add a `.Recover` method and a `.Connectivity` property to the
+`au.com.CodeConstruct.MCTP.Endpoint` interface. `.Recover` takes no arguments,
+produces no result, and returns immediately. `.Connectivity` takes one of two
+values:
+
+- `Available`
+- `Degraded`
```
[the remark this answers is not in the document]
amboar: > What will be default value of .Connectivity?

If the interface has been created on an endpoint object the endpoint is implicitly available, so the default value for `.Connectivity` is `Available`.

> Do the nvmesensor daemon should call .Recover at the first time the Endpoint EID is added to MCTD.Endpoint D-Bus interface to check the ready status of the Endpoint for communication?

I'm not sure I understand the motivation for the question. If the endpoint is added and this is observed through an `InterfacesAdded` signal, then it's inferred that the value for `.Connectivity` is `Available` (as per the default above).

`.Recover` should only be called if an application communicating via an upper-layer protocol over MCTP has concerns that the MCTP endpoint itself is unresponsive. At any other point the application can read `.Connectivity` to determine its current state. Generally this should be implemented by setting a DBus match for a `PropertiesChanged` signal over the `.Connectivity` member, followed by a `GetProperty` to determine its current value.

docs/endpoint-recovery.md
amboar: > Because in this case step 1 and 2 are from out side service, maybe we can add one note to request the service make sure that the device is ready for MCTP interface before call setupEndpoint.

I don't think there should be any mystery to this. Applications that have received some assertion that an MCTP-capable device exists can call `SetupEndpoint` multiple times until it succeeds. It's an error to assume that `SetupEndpoint` will always succeed, just like it's an error to assume that if `SetupEndpoint` succeeds that the endpoint object will always exist. I don't see a need for another mechanism to determine whether the endpoint is ready to communicate via MCTP (unless the device has implementation bugs?).

PR #24, opened 2024-01-12
Add mctp test infrastructure

jk-ozlabs:
This series implements a scriptable test infrastructure for mctpd, mocking both the netlink and AF_MCTP interfaces, to allow arbitrary kernel and external network state.

The majority of the code is a new pytest fixture, which provides small socket interface to an external program (in our case, mctpd), over which the netlink and MCTP communication can be simulated. This allows us to write fairly straightforward tests, including the example in this change:

```python
""" Test the SetupEndpoint dbus call

Using the default system & network ojects, call SetupEndpoint on our mock
endpoint. We expect the dbus call to return the endpoint details, and
the new kernel neighbour and route entries.
"""
async def test_setup_endpoint(dbus, mctpd):
    # shortcuts to the default system/network configuration
    iface = mctpd.system.interfaces[0]
    ep = mctpd.network.endpoints[0]

    # our proxy dbus object for mctpd
    mctp = await mctpd_mctp_obj(dbus)

    # call SetupEndpoint. This will raise an exception on any dbus error.
    (eid, net, path, new) = await mctp.call_setup_endpoint(iface.name, ep.lladdr)

    # ep.eid will be updated (through the Set Endpoint ID message); this
    # should match the returned EID
    assert eid == ep.eid

    # we should have a neighbour for the new endpoint
    assert len(mctpd.system.neighbours) == 1
    neigh = mctpd.system.neighbours[0]
    assert neigh.lladdr == ep.lladdr
    assert neigh.eid == ep.eid

    # we should have a route for the new endpoint too
    assert len(mctpd.system.routes) == 2
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,800 @@
+
+import array
+import trio
+import socket
+import struct
+import sys
+import os
+import uuid
+import errno
+from collections import namedtuple
+
+from pyroute2 import netlink
+from pyroute2.netlink import rtnl
+
+import pytest
+import asyncdbus
+
+AF_NETLINK = 16
+AF_MCTP = 45
+ARPHRD_MCTP = 290
+IFLA_MCTP_NET = 1
+
+MAX_SOCKADDR_SIZE = 56
+
+# can be serialised into a NLMSG_ERROR
+class NetlinkError(Exception):
+    def __init__(self, errno, msg = None):
+        self.errno = errno
+        self.msg = msg
+
+    def to_nlmsg(self, seq = 0):
+        resp = netlink.nlmsgerr()
+        resp['header']['sequence_number'] = seq
+        resp['header']['pid'] = 0
+        resp['header']['type'] = netlink.NLMSG_ERROR
+        resp['error'] = -self.errno
+        if self.msg:
+            resp['attrs'] = [['NLMSGERR_ATTR_MSG', self.msg]]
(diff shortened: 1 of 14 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,793 @@
...
+
+    return Sysnet(system, network)
+
+@pytest.fixture
+async def dbus():
+    async with asyncdbus.MessageBus().connect() as bus:
+        yield bus
+
+@pytest.fixture
+async def mctpd(request, nursery, dbus, sysnet):
```
jk-ozlabs, line 789: hm, we probably don't need the `request` fixture here...

tests/conftest.py
```diff
@@ -0,0 +1,793 @@
...
+    async def start_mctpd(self, nursery):
+        nursery.start_soon(self.handle_control, nursery)
+        self.proc = await nursery.start(self.mctpd_proc, nursery)
+
+    def stop_mctpd(self):
+        if self.proc:
+            self.proc.terminate()
+
+    async def mctpd_proc(self, nursery, task_status = trio.TASK_STATUS_IGNORED):
+        # set up the signal listener
```
jk-ozlabs, line 731: This could probably do with some commenting on why we're interacting with dbus

jk-ozlabs:
Added @amboar 's patch for coverage data.

PR #25, opened 2024-01-12
Endpoint recovery fixes

amboar:
Here are a couple of fixes that were identified with slightly broader testing.

I've exercised these on an AST2600 EVB against a Micron 7450 drive. The drive supports `Get Endpoint UUID` but returns the nil UUID, which isn't terribly helpful, but we have the build feature to deal with it. The three recovery paths have been tested with:

1. Setting up the device's endpoint, removing device power, and leaving it unpowered until recovery fails
2. Building with `-Dunsafe-recover-nil-uuid=false`, setting up the device's endpoint, power-cycling the device, and then invoking a recovery. The original endpoint object was removed, and a new endpoint added, observed via `busctl monitor`
3. Building with `-Dunsafe-recover-nil-uuid=true`, setting up the device's endpoint, power-cycling the device, and then invoking a recovery. The original endpoint object remained, with a signal emitted for `Connectivity` returning to `Available`

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -2281,27 +2281,17 @@ peer_endpoint_recover(sd_event_source *s, uint64_t usec, void *userdata)
 	 */
 	if (peer->recovery.eid != peer->eid) {
 		static const uint8_t nil_uuid[16] = {0};
+		bool uuid_matches_peer = false;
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
LGTM, thanks!

PR #27, opened 2024-01-17
tests: mctpd endpoint recovery

amboar:
There are several paths of interest through recovery:

1. The device is immediately responsive despite the `Recover` invocation
2. The device fails to respond prior to Treclaim
3. The device is initially unresponsive but recovers before Treclaim
4. The device has been exchanged for another

Add test cases for all four.


jk-ozlabs:
Looks great, thanks!

PR #28, opened 2024-01-18
Use Github Actions for CI on pull requests

amboar:


include/linux/mctp.h
jk-ozlabs, line 1: We already have compat definitions in `src/mctp.h`, should be be pulling those in somewhere?

amboar: Ah I missed that, let me understand how that's used.

meson.build
```diff
@@ -80,12 +81,46 @@ configure_file(
...
+    # `sd_bus_default()` can be influenced to select the user session by setting
+    # `DBUS_STARTER_BUS_TYPE=user` in the environment. At this point with a
+    # naive approach we run up against 2, where `dbus-run-session` sanitises
+    # that configuration away
+    #
+    # Invoke pytest via a shell script under `dbus-run-session` so we can
+    # override the sanitation of `DBUS_STARTER_BUS_TYPE`, ensuring `test-mctpd`
+    # connects to the isolated session bus prepared by `dbus-run-session`.
+    pytest = find_program('pytest', required: tests)
+    script = 'export DBUS_STARTER_BUS_TYPE=user ; @0@ --tap'.format(pytest.full_path())
```
jk-ozlabs, line 118: Could we drop the extra `/bin/sh` step by setting the `env` of the `test()` call below?

jk-ozlabs: hm, apparently not:

dbus-run-session(1):

> The variables DBUS_SESSION_BUS_PID, DBUS_SESSION_BUS_WINDOWID, DBUS_STARTER_BUS_TYPE and DBUS_STARTER_ADDRESS are removed from the environment, if present.

jk-ozlabs: ... might help if I read the comment above properly :D

amboar: Can you unpack that thought a little more?

Does the suggestion account for the commentary between 105-116?

amboar: haha, comment race!

PR #29, opened 2024-01-22
Type support

admiyo:
The type code had been hard-coded to 0, which is for MCTP command messages.  However, SPDM requires a type code of 5.  this adds a command line parameter to set the type code.  It leaves the default as 0 if the type parameter is absent.

The second commit removes the error message that comes  back if the request and response are not parallel in size.  This check is only relevant for echo messages, and does not apply for MCTP command or SPDM messages.


admiyo:
I see now that this pretty much echos the "generic" MR which is better documented, and has some feed back.  I will merge in this effort with that one.

PR #33, opened 2024-02-06
Improve checks for MCTP control protocol messages

jk-ozlabs:
This series implements more checking on MCTP control protocol responses: that they have the correct opcode and IID values from the request.

In doing so, we use a common error checking path.


jk-ozlabs:
I have some improvements in mind for this, converting to draft.

PR #34, opened 2024-02-06
mctpd: use new Bus Owner name of `au.com.CodeConstruct.MCTP`

jk-ozlabs:
We currently use a Bus Owner name of `xyz.openbmc_project.MCTP`, to suit conventions with the existing dbus interface names, and MCTP object path structure.

However, mctpd is not under the OpenBMC umbrella, so we should be using our own namespace instead.

Change to a Bus Owner value of `au.com.CodeConstruct.MCTP`, but keep the old value for compatibility. This will be removed in a future release, once we have confidence that consumers are using the new name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 3. The `tests` option has changed type from `feature` to `boolean`. Tests are
    enabled by default.
 
+4. Use a dbus owner name of au.com.CodeConstruct.MCTP. We still register the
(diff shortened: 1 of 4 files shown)


amboar (approved):

jk-ozlabs:
Per some conversation on the OpenBMC discord, we may want to use this as an opportunity to implement a more comprehensive (and versioned) dbus object model.

amboar:
(for the record, the conversation happens around here: https://discord.com/channels/775381525260664832/778790638563885086/1240142924381945856)

jk-ozlabs:
Closed by #43

PR #39, opened 2024-05-20
mctpd: Add `mctp/links/<linkName>` D-Bus object

ThuBaNguyen:
1. Create the MCTP Link D-Bus objects for the existing MCTP links at `/xyz/openbmc_project/mctp/links/<linkName>`. When the MCTP links is removed from/added to the system, the D-Bus object will also be removed/added.
2. Create the au.com.CodeConstruct.MCTP.Link D-Bus interface for MCTP Link D-Bus objects. The interface includes the `Role` property which reports BMC roles in the link. The possible value of `Role` are `BusOwner`, `Endpoint` and `Unknown`.
3. Because the BMC `Role` in the MCTP link is fixed. The `Role` property is changeable value but it can only be changed when the current configured value is `Unknown`.

Ex:
```
~# busctl tree xyz.openbmc_project.MCTP
`- /xyz
  `- /xyz/openbmc_project
    `- /xyz/openbmc_project/mctp
      |- /xyz/openbmc_project/mctp/1
      | `- /xyz/openbmc_project/mctp/1/8
      `- /xyz/openbmc_project/mctp/links
        |- /xyz/openbmc_project/mctp/links/lo
        `- /xyz/openbmc_project/mctp/links/mctpi2c3

~# busctl introspect xyz.openbmc_project.MCTP /xyz/openbmc_project/mctp/links
NAME                                TYPE      SIGNATURE RESULT/VALUE FLAGS
org.freedesktop.DBus.Introspectable interface -         -            -
.Introspect                         method    -         s            -
org.freedesktop.DBus.Peer           interface -         -            -
.GetMachineId                       method    -         s            -
.Ping                               method    -         -            -
org.freedesktop.DBus.Properties     interface -         -            -
.Get                                method    ss        v            -
.GetAll                             method    s         a{sv}        -
.Set                                method    ssv       -            -
.PropertiesChanged                  signal    sa{sv}as  -            -

~# busctl introspect xyz.openbmc_project.MCTP /xyz/openbmc_project/mctp/links/mctpi2c3
NAME                                TYPE      SIGNATURE RESULT/VALUE FLAGS
au.com.CodeConstruct.MCTP.Link      interface -         -            -
.Role                               property  s         "Unknown"    emits-change writable
org.freedesktop.DBus.Introspectable interface -         -            -
.Introspect                         method    -         s            -
org.freedesktop.DBus.Peer           interface -         -            -
.GetMachineId                       method    -         s            -
.Ping                               method    -         -            -
org.freedesktop.DBus.Properties     interface -         -            -
.Get                                method    ss        v            -
.GetAll                             method    s         a{sv}        -
.Set                                method    ssv       -            -
.PropertiesChanged                  signal    sa{sv}as  -            -

~# busctl set-property xyz.openbmc_project.MCTP /xyz/openbmc_project/mctp/links/mctpi2c3 au.com.CodeConstruct.MCTP.Link Role s BusOwner

~# busctl set-property xyz.openbmc_project.MCTP /xyz/openbmc_project/mctp/links/mctpi2c3 au.com.CodeConstruct.MCTP.Link Role
s "BusOwner"
```

### DIFF
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -2,20 +2,28 @@
 
 ## D-Bus
 
-`mctpd` provides a D-Bus path of `/xyz/openbmc_project/mctp`. For each known MCTP endpoint, `mctpd`
-will populate an object `/xyz/openbmc_project/mctp/<NetworkId>/<EID>`. The objects have interface
-`xyz.openbmc_project.MCTP.Endpoint`, as per 
-[OpenBMC documentation](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/yaml/xyz/openbmc_project/MCTP).
+`mctpd` provides a D-Bus service named `au.com.codeconstruct.MCTP1`, and a base
+object path of `/au/com/codeconstruct/mctp1`. For each known MCTP endpoint,
+`mctpd` will populate an object at
+`/au/com/codeconstruct/mctp1/networks/<NetworkId>/endpoints/<EID>`. The objects have
+interface `xyz.openbmc_project.MCTP.Endpoint`, as per [OpenBMC
+documentation](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/yaml/xyz/openbmc_project/MCTP).
 
-As well as the standard interfaces, `mctpd` provides methods to add and configure MCTP endpoints.
-These are provided by the `au.com.CodeConstruct.MCTP` D-Bus interface.
+As well as those standard interfaces, `mctpd` provides methods to add and
+configure MCTP endpoints. These are provided by the `au.com.codeconstruct.MCTP1`
+D-Bus interface.
 
-### `.SetupEndpoint`
+## Bus-owner methods: `au.com.codeconstruct.MCTP.BusOwner1` interface
 
-This method is the normal method used to add a MCTP endpoint.
-The endpoint is identified by MCTP network interface, and physical address.
-`mctpd` will query for the endpoint's current EID, and assign an EID to the endpoint if needed.
-`mctpd` will add local MCTP routes and neighbour table entries for endpoints as they are added.
+This interface exposes bus-owner level functions.
+
+### `.SetupEndpoint`: `say` → `yisb`
+
+This method is the normal method used to add a MCTP endpoint. The endpoint is
+identified by MCTP network interface, and physical address. `mctpd` will query
+for the endpoint's current EID, and assign an EID to the endpoint if needed.
+`mctpd` will add local MCTP routes and neighbour table entries for endpoints as
+they are added.
 
 `SetupEndpoint <interface name> <hwaddr>`
 
@@ -29,23 +37,27 @@ new  (bool) - true if a new EID was assigned
 
 `<interface name>` is an interface such as `mctpi2c6`.
 
-`<hwaddr>` depends on the transport type - for i2c it is a 1 byte client address (7-bit, the same as other Linux tools like `i2cdetect`).
+`<hwaddr>` depends on the transport type - for i2c it is a 1 byte client address
+(7-bit, the same as other Linux tools like `i2cdetect`).
 
 
 An example:
 
 ```shell
-busctl call xyz.openbmc_project.MCTP /xyz/openbmc_project/mctp \
-        au.com.CodeConstruct.MCTP SetupEndpoint say mctpi2c6 1 0x1d
+busctl call au.com.codeconstruct.MCTP1 \
+    /au/com/codeconstruct/mctp1 \
+    au.com.codeconstruct.MCTP1 \
+    SetupEndpoint say mctpi2c6 1 0x1d
 ```
 `1` is the length of the hwaddr array.
 
-### `.AssignEndpoint`
+### `.AssignEndpoint`: `say` → `yisb`
 
-Similar to SetupEndpoint, but will always assign an EID rather than querying for existing ones.
-Will return `new = false` when an endpoint is already known to `mctpd`.
+Similar to SetupEndpoint, but will always assign an EID rather than querying for
+existing ones. Will return `new = false` when an endpoint is already known to
+`mctpd`.
 
-### `.AssignEndpointStatic`
+### `.AssignEndpointStatic`: `sayy` → `yisb`
 
 Similar to AssignEndpoint, but takes an additional EID argument:
 
@@ -59,35 +71,36 @@ This call will fail if the endpoint already has an EID, and that EID is
 different from `static-EID`, or if `static-EID` is already assigned to another
 endpoint.
 
-### `.LearnEndpoint`
+### `.LearnEndpoint`: `say` → `yisb`
 
-Like SetupEndpoint but will not assign EIDs, will only query endpoints for a current EID.
-The `new` return value is set to `false` for an already known endpoint, or `true` when an
-endpoint's EID is newly discovered.
+Like SetupEndpoint but will not assign EIDs, will only query endpoints for a
+current EID. The `new` return value is set to `false` for an already known
+endpoint, or `true` when an endpoint's EID is newly discovered.
 
-## Endpoint Methods
+## Endpoint methods: the `au.com.codeconstruct.MCTP.Endpoint1` interface
 
-Each endpoint object has methods to configure it, with `au.com.CodeConstruct.MCTP.Endpoint`
-interface on each endpoint.
+Each endpoint object has methods to configure it, through the
+`au.com.codeconstruct.MCTP.Endpoint1` interface on each endpoint.
 
-## `.SetMTU`
+## `.SetMTU`: `u`
 
-Sets the MTU (maximum transmission unit) on the route for that endpoint. This must be within
-the MTU range allowed for the network device. For i2c that is [68, 254].
+Sets the MTU (maximum transmission unit) on the route for that endpoint. This
+must be within the MTU range allowed for the network device. For i2c that is
+[68, 254].
 
-If a route-specific MTU has not been set (or set to 0), Linux will use the per-interface
-MTU, configurable with `mctp link set <ifname> mtu <value>`.
+If a route-specific MTU has not been set (or set to 0), Linux will use the
+per-interface MTU, configurable with `mctp link set <ifname> mtu <value>`.
 
 An example, setting MTU of 80:
 
 ```shell
-busctl call xyz.openbmc_project.MCTP  /xyz/openbmc_project/mctp/1/11 \
-        au.com.CodeConstruct.MCTP.Endpoint  SetMTU u 80
+busctl call au.com.codeconstruct.MCTP1 \
+    /au/com/codeconstruct/mctp1/networks/1/endpoints/11 \
+    au.com.codeconstruct.MCTP.Endpoint1 \
+    SetMTU u 80
 ```
 
 ## `.Remove`
 
 Removes the MCTP endpoint from `mctpd`, and deletes routes and neighbour entries.
-
-
 
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -104,3 +104,17 @@ busctl call au.com.codeconstruct.MCTP1 \
 
 Removes the MCTP endpoint from `mctpd`, and deletes routes and neighbour entries.
 
+## D-Bus /au/com/codeconstruct/mctp1/interfaces/<name>
+
+`mctpd` provides a D-Bus path of `/au/com/codeconstruct/mctp1/interfaces`.
+For each known MCTP interfaces, `mctpd` will populate an D-Bus object
+`/au/com/codeconstruct/mctp1/interfaces/<name>`. The D-Bus objects have
+interface `au.com.CodeConstruct.MCTP.Link1`.
+The D-Bus interface includes the `Role` property which reports BMC roles
+in the link. The possible value of `Role` are `BusOwner`, `Endpoint` and
+`Unknown`. The `Role` property is a changeable value but it can only be
+changed when the current configured value is `Unknown` because the BMC
+`Role` in the MCTP link is specific depend on the system.
+The D-Bus `/au/com/codeconstruct/mctp1/interfaces/<name>` objects also
+includes an au.com.codeconstruct.MCTP.BusOwner1 which exposes bus-owner
+level functions.
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -13,7 +13,7 @@ As well as those standard interfaces, `mctpd` provides methods to add and
 configure MCTP endpoints. These are provided by the `au.com.codeconstruct.MCTP1`
 D-Bus interface.
 
-## Bus-owner methods: `au.com.codeconstruct.MCTP.BusOwner1` interface
+## Bus-owner methods: `au.com.codeconstruct.MCTP.BusOwner1.DRAFT` interface
 
 This interface exposes bus-owner level functions.
 
--- a/src/mctp-netlink.c
+++ b/src/mctp-netlink.c
@@ -28,6 +28,8 @@ struct linkmap_entry {
 
 	mctp_eid_t *local_eids;
 	size_t num_local;
+
+	void	*userdata;
 };
 
 struct mctp_nl {
@@ -242,7 +244,7 @@ static void fill_eid_changes(const struct linkmap_entry *oe,
 }
 
 static void fill_link_changes(const struct linkmap_entry *old, size_t old_count,
-	const struct linkmap_entry *new, size_t new_count,
+	struct linkmap_entry *new, size_t new_count,
 	mctp_nl_change **changes, size_t *num_changes) {
 
 	size_t siz = 0;
@@ -250,7 +252,7 @@ static void fill_link_changes(const struct linkmap_entry *old, size_t old_count,
 	// iterate and match old/new interface lists
 	for (size_t o = 0, n = 0; o < old_count || n < new_count; ) {
 		const struct linkmap_entry *oe = &old[o];
-		const struct linkmap_entry *ne = &new[n];
+		struct linkmap_entry *ne = &new[n];
 		mctp_nl_change *ch = NULL;
 
 		if (o >= old_count)
@@ -261,6 +263,7 @@ static void fill_link_changes(const struct linkmap_entry *old, size_t old_count,
 
 		if (oe && ne && oe->ifindex == ne->ifindex) {
 			// Same link.
+			ne->userdata = oe->userdata;
 			if (oe->net == ne->net) {
 				// Same net. Check for eid changes.
 				fill_eid_changes(oe,
@@ -280,6 +283,7 @@ static void fill_link_changes(const struct linkmap_entry *old, size_t old_count,
 				ch->op = MCTP_NL_CHANGE_NET;
 				ch->ifindex = ne->ifindex;
 				ch->old_net = oe->net;
+				ch->link_userdata = oe->userdata;
 			}
 
 			if (oe->up != ne->up) {
@@ -287,6 +291,7 @@ static void fill_link_changes(const struct linkmap_entry *old, size_t old_count,
 				ch->op = MCTP_NL_CHANGE_UP;
 				ch->ifindex = ne->ifindex;
 				ch->old_up = oe->up;
+				ch->link_userdata = oe->userdata;
 			}
 			o++;
 			n++;
@@ -309,6 +314,7 @@ static void fill_link_changes(const struct linkmap_entry *old, size_t old_count,
 			ch->op = MCTP_NL_DEL_LINK;
 			ch->ifindex = oe->ifindex;
 			ch->old_net = oe->net;
+			ch->link_userdata = oe->userdata;
 			o++;
 		}
 	}
@@ -927,6 +933,23 @@ int mctp_nl_net_byindex(const mctp_nl *nl, int index)
 	if (entry)
 		return entry->net;
 	return 0;
+}
+
+int mctp_nl_set_link_userdata(mctp_nl *nl, int ifindex, void *userdata)
+{
+	struct linkmap_entry *entry = entry_byindex(nl, ifindex);
+	if (!entry)
+		return -1;
+
+	entry->userdata = userdata;
+	return 0;
+}
+
+void *mctp_nl_get_link_userdata(const mctp_nl *nl, int ifindex)
+{
+	struct linkmap_entry *entry = entry_byindex(nl, ifindex);
+
+	return entry ? entry->userdata : NULL;
 }
 
 bool mctp_nl_up_byindex(const mctp_nl *nl, int index)
--- a/src/mctp-netlink.c
+++ b/src/mctp-netlink.c
@@ -952,6 +952,19 @@ void *mctp_nl_get_link_userdata(const mctp_nl *nl, int ifindex)
 	return entry ? entry->userdata : NULL;
 }
 
+void *mctp_nl_get_link_userdata_byname(const mctp_nl *nl, const char *ifname)
+{
+       size_t i;
+
+       for (i = 0; i < nl->linkmap_count; i++) {
+               struct linkmap_entry *entry = &nl->linkmap[i];
+               if (!strcmp(entry->ifname, ifname))
+                       return entry->userdata;
+       }
+
+       return NULL;
+}
+
 bool mctp_nl_up_byindex(const mctp_nl *nl, int index)
 {
 	struct linkmap_entry *entry = entry_byindex(nl, index);
--- a/src/mctp-netlink.c
+++ b/src/mctp-netlink.c
@@ -47,11 +47,6 @@ struct mctp_nl {
 	bool quiet_eexist;
 };
 
-#define max(a, b) ((a) > (b) ? (a) : (b))
-#define min(a, b) ((a) < (b) ? (a) : (b))
-
-#define ARRAY_SIZE(a) (sizeof(a)/sizeof((a)[0]))
-
 static int fill_local_addrs(mctp_nl *nl);
 static int fill_linkmap(mctp_nl *nl);
 static void sort_linkmap(mctp_nl *nl);
--- a/src/mctp-netlink.h
+++ b/src/mctp-netlink.h
@@ -84,6 +84,8 @@ int mctp_nl_set_link_userdata(mctp_nl *nl, int ifindex, void *userdata);
 
 /* Returns NULL if the link does not exist */
 void *mctp_nl_get_link_userdata(const mctp_nl *nl, int ifindex);
+/* Returns NULL if the link does not exist */
+void *mctp_nl_get_link_userdata_byname(const mctp_nl *nl, const char *ifname);
 
 /* MCTP route helper */
 int mctp_nl_route_add(struct mctp_nl *nl, uint8_t eid, const char* ifname,
--- a/src/mctp-netlink.h
+++ b/src/mctp-netlink.h
@@ -31,6 +31,10 @@ struct mctp_nl_change {
 
 	// Filled for CHANGE_UP
 	bool old_up;
+
+	// If userdata is present on the link, it is passed here. Populated for
+	// link change events (DEL_LINK, CHANGE_NET, CHANGE_UP).
+	void *link_userdata;
 };
 typedef struct mctp_nl_change mctp_nl_change;
 
@@ -68,6 +72,18 @@ void mctp_nl_linkmap_dump(const mctp_nl *nl);
 int *mctp_nl_net_list(const mctp_nl *nl, size_t *ret_num_nets);
 /* Returns an allocated list of ifindex, caller to free */
 int *mctp_nl_if_list(const mctp_nl *nl, size_t *ret_num_if);
+
+/* Get/set userdata for a link. The userdata is attached to a link
+ * with index @ifindex. Userdata will also be populated into
+ * struct mctp_nl_change->userdata, and would typically be freed on
+ * MCTP_NL_DEL_LINK events
+ *
+ * Returns non-zero if the link does not exist.
+ */
+int mctp_nl_set_link_userdata(mctp_nl *nl, int ifindex, void *userdata);
+
+/* Returns NULL if the link does not exist */
+void *mctp_nl_get_link_userdata(const mctp_nl *nl, int ifindex);
 
 /* MCTP route helper */
 int mctp_nl_route_add(struct mctp_nl *nl, uint8_t eid, const char* ifname,
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -42,12 +42,14 @@
 
 #define MCTP_DBUS_PATH "/au/com/codeconstruct/mctp1"
 #define MCTP_DBUS_PATH_NETWORKS "/au/com/codeconstruct/mctp1/networks"
+#define MCTP_DBUS_PATH_LINKS "/au/com/codeconstruct/mctp1/interfaces"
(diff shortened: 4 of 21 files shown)


jk-ozlabs:
Thanks!

We'll need to check whether this is compatible with the existing layout where the top-level `/mctp/` can host a non-network child object. If not, we will have to move those to `endpoints/`.

jk-ozlabs:
I started writing some notes on the interface changes, but that expanded beyond just this change, so they're now in #40.

However, I'd rather not drag out this change too long, so already have progress on implementing the changes in #40, which should be quick to adapt to most design changes that may arise there.

So, I'll add a couple of review comments on this, just for top-level API things for now; we can cover implementation details soon. It's my goal that we would then rebase this change on to the dbus api rework that comes out of #40.

docs/mctpd.md
```diff
@@ -89,5 +89,13 @@ busctl call xyz.openbmc_project.MCTP  /xyz/openbmc_project/mctp/1/11 \
 
 Removes the MCTP endpoint from `mctpd`, and deletes routes and neighbour entries.
 
+## D-Bus /xyz/openbmc_project/mctp/links/<InterfaceName>
```
jk-ozlabs, line 92: Some bikeshedding that may come out of #40: `links` or `interfaces` (or something else?)

src/mctp-netlink.h
```diff
@@ -3,12 +3,53 @@
...
+	} role;
+
+	mctp_eid_t *local_eids;
+	size_t num_local;
+};
+
+#define MCTP_ENTITY_ROLE_COUNT 3
+static const char * const role_name[] = {
+	[MCTP_ENTITY_ROLE_BUS_OWNER] = "BusOwner",
+	[MCTP_ENTITY_ROLE_ENDPOINT] = "EndPoint",
```
jk-ozlabs, line 30: "endpoint" is one word, so `Endpoint` for this enumeration.

src/mctpd.c
```diff
@@ -2725,6 +2757,144 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+		goto out;
+	}
+	dfree(tmpstr);
+	dfree(link_name);
+
+	if (strcmp(property, "Role") == 0) {
+		if (lm->role != MCTP_ENTITY_ROLE_UNKNOWN)
+		{
+			sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+						"Can't change the none \"Unknown\" BMC Role.");
```
jk-ozlabs, line 2856: This error message is a little confusing; maybe "Role is already set"?

src/mctpd.c
```diff
@@ -2883,6 +3118,26 @@ static char* net_path(int net)
...
+	size_t l;
+	char *buf = NULL;
+
+	l = strlen(MCTP_DBUS_PATH_LINKS) + 30;
+	buf = malloc(l);
+	if (!buf) {
+		return NULL;
+	}
+	/* can't use sd_bus_path_encode_many() since it escapes
+	   leading digits */
```
jk-ozlabs, line 3132: will we have leading digits?

(I'm aware that this is copied from the peer_find function, but we have different naming semantics here)

jk-ozlabs (commented):
Mostly looks good, we'll just need to synchronise with the dbus API changes.

src/mctpd.c
mkj: Interface names are allowed leading digits. eg `ip tuntap add name 123tun mode tun`

docs/mctpd.md
ThuBaNguyen: I updated code to follow the propose #40.

src/mctpd.c
```diff
@@ -2734,19 +2818,150 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+		goto out;
+	}
+
+	lm = find_link_by_name(ctx, link_name);
+	if (!lm) {
+		rc = -ENOENT;
+		goto out;
+	}
+	dfree(tmpstr);
+	dfree(link_name);
```
jk-ozlabs, line 2851: do these need to be deferred?

jk-ozlabs (commented):
Mostly looks good, thanks!

I have a few comments inline, but a lot of those are more questions to improve my understanding of your approach here.

src/mctpd.c
```diff
@@ -153,13 +156,19 @@ struct ctx {
 	// Whether we are running as the bus owner
 	bool bus_owner;
 
+	// default BMC role in all active MCTP interfaces
+	size_t default_bmc_role;
```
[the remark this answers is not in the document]
ThuBaNguyen: default_bmc_role has three possible values while bus_owner only have two. Can I change type of bus_owner?

src/mctpd.c
```diff
@@ -2734,19 +2818,150 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+					"Unknown Role. Only Support BusOwner/EndPoint.");
+			goto out;
+		}
+		for (i = 0; i < MCTP_ENTITY_ROLE_COUNT; i++)
+		{
+			if (strcmp(state, role_name[i]) == 0)
+			{
+				found = true;
+				break;
+			}
```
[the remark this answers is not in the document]
jk-ozlabs: > If you like, I can introduce that in an earlier change for `parse_config_mode`, and you'd just add the `dbus_val` to that.

ok, done: https://github.com/CodeConstruct/mctp/commit/2830d8afc542fdbf46a4592125ff0bb1fc5430e2

src/mctpd.c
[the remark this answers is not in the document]
jk-ozlabs: yeah, I think that would suit best. We can later rename this to the `default_` setting, once we can have different roles between links (and then have the actual role setting per-link).

src/mctpd.c
ThuBaNguyen: Can you give me some guideline when dfree() is need in mctpd.c?
I just copy/parse from other function.

src/mctpd.c
```diff
@@ -2734,19 +2818,150 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+
+
+	if (strcmp(property, "Role") == 0) {
+		if (lm->role != MCTP_ENTITY_ROLE_UNKNOWN)
+		{
+			sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+						"Role is already set.");
+			rc = -ENOENT;
+			goto out;
+		}
```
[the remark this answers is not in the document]
ThuBaNguyen: No I don't change  indentation styles. Still use tab.

src/mctpd.c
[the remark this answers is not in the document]
ThuBaNguyen: I updated the code to follow the change in https://github.com/CodeConstruct/mctp/commit/2830d8afc542fdbf46a4592125ff0bb1fc5430e2

src/mctp-netlink.h
```diff
@@ -3,12 +3,54 @@
...
+	size_t	linkmap_alloc;
+	bool	verbose;
+	bool	publish_links;
+
+	// allows callers to silence printf of EEXIST returns.
+	// TODO: this is a workaround, if more are required we should
+	// rework how error messages are returned to callers.
+	bool quiet_eexist;
+};
+
```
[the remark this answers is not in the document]
ThuBaNguyen: Done.

src/mctpd.c
```diff
@@ -153,13 +156,19 @@ struct ctx {
...
+
 	// An allocated array of peers, changes address (reallocated) during runtime
 	peer *peers;
 	size_t size_peers;
 
 	struct net_det *nets;
 	size_t num_nets;
 
+	struct linkmap_entry *linkmap;
+	size_t	linkmap_count;
```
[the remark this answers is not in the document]
ThuBaNguyen: `nl->linkmap` field is reinitialize in `get/set property` callback or event callback. That why I can't use `nl->linkmap`. (I tried and figure out that before add new field to `ctx`)

src/mctpd.c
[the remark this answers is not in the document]
ThuBaNguyen: I updated the code. Please let me know your idea.

src/mctpd.c
```diff
@@ -2734,19 +2818,150 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
-                                 const char *interface,
-                                 const char *property,
-                                 sd_bus_message *value,
-                                 void *userdata,
-                                 sd_bus_error *ret_error)
+								 const char *interface,
+								 const char *property,
+								 sd_bus_message *value,
+								 void *userdata,
+								 sd_bus_error *ret_error)
```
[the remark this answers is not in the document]
ThuBaNguyen: I removed this changing.

src/mctpd.c
```diff
@@ -2973,14 +3322,39 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
...
+		num_nodes++;
+		for (size_t t = 0; t < 256; t++) {
+			if (ctx->nets[i].peeridx[t] != -1)
+			{
+				// .../mctp1/networks/<NetID>/endpoints object
+				num_nodes++;
+				break;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ThuBaNguyen: I created one commit https://github.com/CodeConstruct/mctp/pull/39/commits/61b7e03cc6f7cac0b9c899ee96149a5c651c576e

docs/mctpd.md
```diff
@@ -104,3 +104,11 @@ busctl call au.com.codeconstruct.MCTP1 \
...
 Removes the MCTP endpoint from `mctpd`, and deletes routes and neighbour entries.
 
+## D-Bus /xyz/openbmc_project/mctp1/interfaces/<name>
+
+`mctpd` provides a D-Bus path of `/xyz/openbmc_project/mctp1/interfaces`.
+For each known MCTP interfaces, `mctpd` will populate an object `/xyz/openbmc_project/mctp1/interfaces/<name>`. The D-Bus objects have interface `au.com.CodeConstruct.MCTP.Link1`. The D-Bus interface includes the `Role` property which reports
+BMC roles in the link. The possible value of `Role` are `BusOwner`, `Endpoint` and `Unknown`.
+Because the BMC `Role` in the MCTP link is fixed. The `Role` property is changeable value but it can only be changed
+when the current configured value is `Unknown`.
+The D-Bus `mctp1/interfaces/<name>` objects also includes an au.com.codeconstruct.MCTP.BusOwner1 which exposes bus-owner level functions.
```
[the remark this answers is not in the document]
ThuBaNguyen: Do you mean change `bus-owner` to `BusOwner`?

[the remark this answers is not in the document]
jk-ozlabs: no, these additions still refer to the `xyz.openbmc_project` namespace, which no longer exists here.

src/mctpd.c
jk-ozlabs: `dfree` is a utility for a "deferred free" of the pointer; the pointer is valid until we return to the dbus event handler, and will be freed during that processing.

In this case, it looks like we don't need those pointers for anything else, so we can free them immediately with regular `free`.

ThuBaNguyen: I updated the code.

docs/mctpd.md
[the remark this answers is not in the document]
ThuBaNguyen: Updated.

src/mctpd.c
```diff
@@ -97,14 +93,15 @@ struct role {
 static const struct role roles[] = {
 	[ENDPOINT_ROLE_UNKNOWN] = {
 		.role = ENDPOINT_ROLE_UNKNOWN,
+		.conf_val = "Unknown",
 	},
 	[ENDPOINT_ROLE_BUS_OWNER] = {
 		.role = ENDPOINT_ROLE_BUS_OWNER,
-		.conf_val = "bus-owner",
+		.conf_val = "BusOwner",
```
jk-ozlabs, line 100: I was more intending that you add a separate `const char *dbus_val` to `struct role`. You would then use this for the dbus-value parsing.

src/mctpd.c
```diff
@@ -176,8 +173,8 @@ struct ctx {
 	// Second instance for sending mctp socket requests. State is unused.
 	mctp_nl *nl_query;
 
-	// Whether we are running as the bus owner
-	bool bus_owner;
+	// Default BMC role in All of MCTP medium interface
+	size_t default_role;
```
jk-ozlabs, line 177: `enum endpoint_role`?

src/mctpd.c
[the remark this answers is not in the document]
jk-ozlabs: you're flipping between:

```c
    if (thing) {
        stuff;
    } else {
        other_stuff;
    }
```

and

```c
    if (thing)
    {
        stuff;
    }
    else
    {
        other_stuff;
    }
```

I'd prefer the former if possible.

(there *may* be some inconsistency in the code already, but I'd like to avoid adding more)

src/mctpd.c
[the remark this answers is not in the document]
jk-ozlabs: looks good!

src/mctpd.c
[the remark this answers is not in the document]
jk-ozlabs: OK, I'm going to have to take some time to see where the linkmaps may diverge, and/or see if we can unify these. Will update shortly.

src/mctpd.c
ThuBaNguyen: Because the possible value of `Mode` in configuration `/conf/mctpd.conf` is change to `BusOwner`, `Endpoint` and `Unknown` as the change in mctpd.conf.
Keeping `conf_val` and its` values is unnecessary as my opinion.

src/mctpd.c
ThuBaNguyen: Done.

src/mctpd.c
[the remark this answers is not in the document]
ThuBaNguyen: > you're flipping between:
> 
> ```c
>     if (thing) {
>         stuff;
>     } else {
>         other_stuff;
>     }
> ```
> 
I updated code to use this style which is mostly used by the mctpd code.

src/mctpd.c
jk-ozlabs: > The configureable values of `conf_val` have to be changed also.

Yep, that's the issue. This change will expose dbus-style names which are not convention for a toml-style configuration. We'll need to keep those separate.

src/mctpd.c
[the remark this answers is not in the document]
jk-ozlabs: > `nl->linkmap` field is reinitialize in `get/set property` callback or event callback. That why I can't use `nl->linkmap`.

I'm not sure I fully understand the concern. Could you elaborate a little on why you can't use the existing linkmap here?

The `mctp-netlink` code is intended for exactly this - handling the netlink messages and establishing our databaset of current links, and synchronising it with the kernel's link states.

With this change, you have added *another* database of links, and code to synchronise that with the `mctp-netlink`'s own linkmap.

I would prefer we didn't do this, and keep the mctp-netlink module structure intact (including non-public accessors for the linkmap database). With your header change split out, I see you have added a couple of bits of information to the linkmap:

```diff
 struct linkmap_entry {
 	int	ifindex;
 	char	ifname[IFNAMSIZ+1];
 	int	net;
 	bool 	up;
+	bool	published;
+	enum endpoint_role role;
 
 	mctp_eid_t *local_eids;
 	size_t num_local;
```

If this is the reason for the duplication, then I'd suggest we can do this much more simply by keeping the linkmap private to mctp-netlink, and provide a way to attach mctpd-specific data to the linkmap entries.

Something like:

```diff
diff --git a/src/mctp-netlink.h b/src/mctp-netlink.h
index 31ee6b6..3bf5d73 100644
--- a/src/mctp-netlink.h
+++ b/src/mctp-netlink.h
@@ -31,6 +31,10 @@ struct mctp_nl_change {
 
        // Filled for CHANGE_UP
        bool old_up;
+
+       // If userdata is present on the link, it is passed here. Populated for
+       // link change events (DEL_LINK, CHANGE_NET, CHANGE_UP).
+       void *link_userdata;
 };
 typedef struct mctp_nl_change mctp_nl_change;
 
@@ -69,6 +73,18 @@ int *mctp_nl_net_list(const mctp_nl *nl, size_t *ret_num_nets);
 /* Returns an allocated list of ifindex, caller to free */
 int *mctp_nl_if_list(const mctp_nl *nl, size_t *ret_num_if);
 
+/* Get/set userdata for a link. The userdata is attached to a link
+ * with index @ifindex. Userdata will also be populated into
+ * struct mctp_nl_change->userdata, and would typically be freed on
+ * MCTP_NL_DEL_LINK events
+ *
+ * Returns non-zero if the link does not exist.
+ */
+int mctp_nl_set_userdata(mctp_nl *nl, int ifindex, void *userdata);
+
+/* Returns NULL if the link does not exist */
+void *mctp_nl_get_userdata(const mctp_nl *nl, int ifindex);
+
 /* MCTP route helper */
 int mctp_nl_route_add(struct mctp_nl *nl, uint8_t eid, const char* ifname,
                uint32_t mtu);

```

So, for the mctpd usage of the linkmap, it would have this for the link userdata:

```c
struct link_data {
    bool published;
    enum endpoint_role role;
}
```

Which mctp would set (using `mctp_nl_set_userdata`) on new links, using the default values. For updates, mctpd would do something like:

Then your property set would look something like:

```c
    	struct link_data *linkdata;

        /* ... */

	rc = get_role(state, &role);
	if (rc < 0) {
		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
			"Unknown Role. Only Support BusOwner/EndPoint.");
		rc = -EINVAL;
		goto out;
	}

	ifindex = mctp_nl_ifindex_byname(ctx->nl, link_name);
	if (ifindex < 0) {
		rc = -ENOENT;
		goto out;
	}

	linkdata = mctp_nl_get_userdata(ctx->nl, ifindex);
	if (!linkdata) {
		rc = -ENOENT;
		goto out;
	}

	if (linkdata->role != ENDPOINT_ROLE_UNKNOWN) {
		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
			"Role is already set.");
		rc = -ENOENT;
		goto out;
	}

	linkdata->role = role.role;
```

Would that work?

If so, I have the required changes to the netlink interface at https://github.com/CodeConstruct/mctp/commit/2e09cfce986ee8135e22a9f822fb49978f70e9cf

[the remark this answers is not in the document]
jk-ozlabs: > If so, I have the required changes to the netlink interface at [2e09cfc](https://github.com/CodeConstruct/mctp/commit/2e09cfce986ee8135e22a9f822fb49978f70e9cf)

... now with a small naming change, (`mctp_nl_set_link_userdata` rather than `mctp_nl_set_userdata`): https://github.com/CodeConstruct/mctp/commit/44cc89bf51ac5ae25db5764c0d4472e252065b4f

This is in a new `dev/linkmap-userdata` branch.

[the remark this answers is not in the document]
ThuBaNguyen: I pinged you on discord.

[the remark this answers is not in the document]
ThuBaNguyen: Done.

src/mctp-netlink.c
jk-ozlabs, line 1: With the link userdata function, you shouldn't need to change this anymore.

src/mctp-netlink.h
```diff
@@ -8,6 +8,12 @@
 
 #include "mctp.h"
 
+enum endpoint_role {
+	ENDPOINT_ROLE_UNKNOWN,
+	ENDPOINT_ROLE_BUS_OWNER,
+	ENDPOINT_ROLE_ENDPOINT,
+};
+
```
jk-ozlabs, line 16: This is specific to the mctpd usage of the netlink code. This should be private to mctpd.c.

src/mctp-netlink.h
```diff
@@ -37,6 +43,11 @@ struct mctp_nl {
 	bool quiet_eexist;
 };
 
+struct link_userdata {
+    bool published;
+    enum endpoint_role role;
+};
+typedef struct link_userdata link_userdata;
```
jk-ozlabs, line 50: Same here - this is specific to the mctpd usage of the netlink code.

src/mctpd.c
```diff
@@ -2760,6 +2828,147 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+	rc = sd_bus_path_decode_many(path, MCTP_DBUS_PATH "/%/%", &tmpstr,
+		&link_name);
+	if (rc <= 0) {
+		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+				"Invalid Object Path");
+		goto out;
+	}
+
+	lmUserData =
+		(link_userdata*)mctp_nl_get_link_userdata_byname(ctx->nl, link_name);
```
jk-ozlabs, line 2856: no need to cast

src/mctpd.c
```diff
@@ -2760,6 +2828,147 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS, " %s lmUserData->role %d lmUserData->published %d", __func__, lmUserData->role, lmUserData->published);
+		goto out;
+	}
+
+	if (lmUserData->published != 0 &&
+		lmUserData->published != 1)
+	{
+		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS, " %s lmUserData->role %d lmUserData->published %d", __func__, lmUserData->role, lmUserData->published);
+		goto out;
+	}
```
jk-ozlabs, line 2874: If we want to check whether the userdata is valid, we should include a magic number in the struct, rather than comparing each field against the entire range of valid values.

However, I'm not sure these checks (or a magic nember check) is required anyway.

src/mctpd.c
```diff
@@ -2760,6 +2828,147 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+	rc = sd_bus_path_decode_many(path, MCTP_DBUS_PATH "/%/%", &tmpstr,
+		&link_name);
+	if (rc <= 0) {
+		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+				"Invalid Object Path");
+		goto out;
+	}
+
+	lmUserData =
+		(link_userdata*)mctp_nl_get_link_userdata_byname(ctx->nl, link_name);
```
jk-ozlabs, line 2917: no need to cast

src/mctpd.c
```diff
@@ -2760,6 +2828,147 @@ static int bus_endpoint_get_prop(sd_bus *bus,
...
+		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS, " %s lmUserData->role %d lmUserData->published %d", __func__, lmUserData->role, lmUserData->published);
+		goto out;
+	}
+
+	if (lmUserData->published != 0 &&
+		lmUserData->published != 1)
+	{
+		sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS, " %s lmUserData->role %d lmUserData->published %d", __func__, lmUserData->role, lmUserData->published);
+		goto out;
+	}
```
jk-ozlabs, line 2936: same as above with the checks here

src/mctpd.c
```diff
@@ -2916,6 +3147,43 @@ static char* root_endpoints_path(int net)
...
+
+	rc = sd_bus_path_decode_many(path, MCTP_DBUS_PATH "/%/%", &tmpstr,
+		&link_name);
+	if (rc == 0)
+		return -ENOENT;
+	if (rc < 0)
+		return rc;
+
+	lmUserData =
+		(link_userdata*)mctp_nl_get_link_userdata_byname(ctx->nl, link_name);
```
jk-ozlabs, line 3173: No need to cast

src/mctpd.c
```diff
@@ -2977,6 +3260,29 @@ static int emit_net_added(ctx *ctx, int net) {
...
+	}
+
+	path = interface_path(ifname);
+	if (path == NULL) {
+		warnx("%s: out of memory", __func__);
+		return -ENOMEM;
+	}
+	rc = sd_bus_emit_object_added(ctx->bus, dfree(path));
+	if (rc < 0)
+		warnx("%s: error emitting, %s", __func__, strerror(-rc));
```
jk-ozlabs, line 3281: Just a tip on this: `warn` (rather than `warnx`) uses the errno value to automatically add the strerror value. So, if you prefer, you could do:

```c
if (rc < 0) {
    errno = -rc;
    warn("%s: error emitting", __func__);
}
```

\- and you would get the error string reported just the same.

src/mctpd.c
```diff
@@ -3035,6 +3365,18 @@ static int mctpd_dbus_enumerate(sd_bus *bus, const char* path,
 		if (ctx->peers[i].published)
 			num_nodes++;
 
+	// .../mctp1/interfaces object
+	num_nodes++;
+
+	// .../mctp1/interface/<name>
+	for (i = 0; i < ctx->nl->linkmap_count; i++) {
```
jk-ozlabs, line 3372: The linkmap definition should be private to netlink.c. Instead of this, use `mctp_nl_if_list` and iterate over the results of that.

src/mctpd.c
```diff
@@ -3477,11 +3858,19 @@ static int add_interface_local(ctx *ctx, int ifindex)
...
 			return rc;
 	}
-
 	eids = mctp_nl_addrs_byindex(ctx->nl, ifindex, &num);
 	for (size_t j = 0; j < num; j++) {
 		add_local_eid(ctx, net, eids[j]);
 	}
+
+	// Add new link if required
+	lmUserData = (link_userdata*)mctp_nl_get_link_userdata(ctx->nl, ifindex);
```
jk-ozlabs, line 3867: No cast needed

src/mctpd.c
```diff
@@ -3512,6 +3901,41 @@ static int add_net(ctx *ctx, int net)
 	return 0;
 }
 
+static int add_interface(ctx *ctx, int ifindex)
+{
+	link_userdata *lmUserData = (link_userdata *)malloc(sizeof(link_userdata));
```
jk-ozlabs, line 3906: No cast needed.

Also, you don't seem to be `free()`ing this. You'll need to do so on the `MCTP_NL_DEL_LINK` notification.

src/mctpd.c
```diff
@@ -3512,6 +3901,41 @@ static int add_net(ctx *ctx, int net)
...
+	if (net <= 0) {
+		warnx("Can't find link index %d\n", ifindex);
+		return -ENOMEM;
+	}
+
+	/* Use the `mode` setting in conf/mctp.conf */
+	lmUserData->role = ctx->default_role;
+	lmUserData->published = true;
+
+	rc = mctp_nl_set_link_userdata(ctx->nl, ifindex, (void*)lmUserData);
```
jk-ozlabs, line 3919: No cast needed

src/mctpd.c
```diff
@@ -3512,6 +3901,41 @@ static int add_net(ctx *ctx, int net)
...
 }
 
+static int add_interface(ctx *ctx, int ifindex)
+{
+	link_userdata *lmUserData = (link_userdata *)malloc(sizeof(link_userdata));
+	int rc;
+
+	int net = mctp_nl_net_byindex(ctx->nl, ifindex);
+	if (net <= 0) {
+		warnx("Can't find link index %d\n", ifindex);
```
jk-ozlabs, line 3911: No newline needed on `warn` / `warnx` / `err` / `errx`.

docs/mctpd.md
```diff
@@ -1,6 +1,6 @@
 # `mctpd`
 
-## D-Bus
+## D-Bus /au/com/codeconstruct/mctp1/<NetworkId>/endpoints/<EID>
```
jk-ozlabs, line 3: This section describes both the root object `.../mctp1` as well as the endpoint objects `.../mctp1/networks/<net>/endpoints/<eid>`, so I don't think we need change this in the section title.

(arguably, we should split that up, but that's a separate change...)

src/mctpd.c
```diff
@@ -895,6 +924,9 @@ static int cb_listen_monitor(sd_event_source *s, int sd, uint32_t revents,
 
 		case MCTP_NL_DEL_LINK:
 		{
+			link_userdata* lmUserData =
```
jk-ozlabs, line 927: the userdata is provided in `c->link_userdata`, you won't need a separate lookup here.

(see https://github.com/CodeConstruct/mctp/commit/44cc89bf51ac5ae25db5764c0d4472e252065b4f#diff-3816167d574db6f7e0cd38ed18a7a4f1ffabe3a057a2421b2c364e7668c5202dR34-R37)

src/mctpd.c
jk-ozlabs: you've resolved this comment, but there are still a bunch of new "\n" in warnx here.

jk-ozlabs (commented):
Looking good! A couple of really minor things, but I think we're almost ready to go here. I'll start some testing on my side too.

jk-ozlabs:
A couple of new warnings have been introduced:

```
[19/23] Compiling C object mctpd.p/src_mctpd.c.o
../src/mctpd.c: In function ‘mctpd_dbus_enumerate’:
../src/mctpd.c:3363:40: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
 3363 |         for (size_t i = 0; i < num_ifs && rc == 0; i++) {
      |                            ~~~~~~~~~~~~^~~~~~~~~~
../src/mctpd.c: In function ‘emit_interface_removed’:
../src/mctpd.c:3301:12: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
 3301 |         if (rc < 0) {
      |            ^
```

but those should be straightforward fixes

ThuBaNguyen:
> A couple of new warnings have been introduced:
> 
> ```
> [19/23] Compiling C object mctpd.p/src_mctpd.c.o
> ../src/mctpd.c: In function ‘mctpd_dbus_enumerate’:
> ../src/mctpd.c:3363:40: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
>  3363 |         for (size_t i = 0; i < num_ifs && rc == 0; i++) {
>       |                            ~~~~~~~~~~~~^~~~~~~~~~
> ../src/mctpd.c: In function ‘emit_interface_removed’:
> ../src/mctpd.c:3301:12: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
>  3301 |         if (rc < 0) {
>       |            ^
> ```
> 
> but those should be straightforward fixes

I fixed it.

jk-ozlabs:
Seems to be one more remaining though:

```
./src/mctpd.c: In function ‘emit_interface_removed’:
../src/mctpd.c:3299:12: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
 3299 |         if (rc < 0) {
      |            ^
```

ThuBaNguyen:
> Seems to be one more remaining though:
> 
> ```
> ./src/mctpd.c: In function ‘emit_interface_removed’:
> ../src/mctpd.c:3299:12: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
>  3299 |         if (rc < 0) {
>       |            ^
> ```

I updated code.

jk-ozlabs:
So this PR got auto-closed on merge of the pre-req dbus-rework branch, but we still do want the change. I have rebased the remaining change and created #44 to continue the interface objects work.

jk-ozlabs:
Ah, seems like the auto-close was because the `dbus-rework` branch was the target for this PR. We want your branch merged into `main` :)

PR #41, opened 2024-06-26
mctpd: fix return value check in control socket reads

jk-ozlabs:
Currently, we're checking the return value of read_message against 0 to detect EOF, but this isn't a length value. This means that any successful read will hit the errx.

Instead, check the actual length, and improve the read_message case where we see an EOF from the recvmsg().

Fixes: ea5c9a3 ("mctpd: Exit on control socket EOF")


jk-ozlabs:
Now with a testcase.

jk-ozlabs:
(merged at https://github.com/CodeConstruct/mctp/commit/36da928bdc3fd3f4af382e079b4997e6d11a363c)

PR #49, opened 2024-08-13
mctpd: fix incorrect response of Get Message Type Support

stevenlee7189:
The response from the "Get Message Type Support" request doesn't include the list of message type numbers; it only returns the message type count.

Request:
00 81 05

Response:
00 00 05 00 01

To fix this issue, adjust the response length to include the message type list. The corrected response should be:

00 00 05 00 01 00


jk-ozlabs:
Good catch, thanks!

(I have a todo to be able to register new message types too)

PR #51, opened 2024-08-30
mctpd: Set initial route MTU to interface minimum

mkj:
This allows better compatibility with devices that have a low initial allowed packet size and require application negotiation to increase that packet size.

Previously the MTU was left at 0, so matching the currently set MTU of the device.

### DIFF
--- a/src/mctp-netlink.c
+++ b/src/mctp-netlink.c
@@ -26,6 +26,9 @@ struct linkmap_entry {
 	int	net;
 	bool 	up;
 
+	uint32_t min_mtu;
+	uint32_t max_mtu;
+
 	mctp_eid_t *local_eids;
 	size_t num_local;
 
@@ -54,7 +57,7 @@ static int fill_linkmap(mctp_nl *nl);
 static void sort_linkmap(mctp_nl *nl);
 static int linkmap_add_entry(mctp_nl *nl, struct ifinfomsg *info,
 		const char *ifname, size_t ifname_len, int net,
-		bool up);
+		bool up, uint32_t min_mtu, uint32_t max_mtu);
 static struct linkmap_entry *entry_byindex(const mctp_nl *nl,
 	int index);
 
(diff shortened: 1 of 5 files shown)


jk-ozlabs:
Couple of minor things:

> Previously the MTU was left at 0, so matching the currently set MTU of the device.

Might be good to clarify this with "link" (or "interface") rather than "device". In the previous paragraph, "device" refers to the remote endpoint.

Could you add a signed-off-by too?

This probably warrants changelog entry as well, as it may trigger application changes.

src/mctp-netlink.c
```diff
@@ -709,21 +711,33 @@ static int parse_getlink_dump(mctp_nl *nl, struct nlmsghdr *nlh, uint32_t len)
...
+
+		if (!mctp_get_rtnlmsg_attr_u32(IFLA_MIN_MTU, rta, rlen, &min_mtu)) {
+			warnx("Missing IFLA_MIN_MTU for %s", ifname);
+			continue;
+		}
+
+		if (!mctp_get_rtnlmsg_attr_u32(IFLA_MAX_MTU, rta, rlen, &max_mtu)) {
+			warnx("Missing IFLA_MAX_MTU for %s", ifname);
+			continue;
+		}
```
jk-ozlabs, line 735: `lo` reports:

```
[{nla_len=8, nla_type=IFLA_MTU}, 65536], [{nla_len=8, nla_type=IFLA_MIN_MTU}, 0], [{nla_len=8, nla_type=IFLA_MAX_MTU}, 0]
```

Do we need any special handling for zero?

mkj: A max MTU of 0 will result in the link MTU being set to 0, the same as currently, so that should be fine. (If properties were optional we could omit min_mtu/max_mtu if they were 0, but it doesn't seem worth adding another interface just for that special case)

PR #54, opened 2024-11-15
Support `LocaEids` property in `au.com.codeconstruct.MCTP.Network1`

ThuBaNguyen:
Add code to show the local EIDs in `EIDs` property in `au.com.CodeConstruct.MCTP.Interface1` D-Bus interface.

```
busctl introspect au.com.codeconstruct.MCTP1 /au/com/codeconstruct/mctp1/interfaces/mctpi2c3
NAME                                 TYPE      SIGNATURE RESULT/VALUE FLAGS
au.com.CodeConstruct.MCTP.Interface1 interface -         -            -
.EIDs                                property  ay        1 8          const
```

### DIFF
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -143,6 +143,27 @@ Like SetupEndpoint but will not assign EIDs, will only query endpoints for a
 current EID. The `new` return value is set to `false` for an already known
 endpoint, or `true` when an endpoint's EID is newly discovered.
 
+## Network objects: `/au/com/codeconstruct/networks/<net>`
+
+These objects represent MCTP networks which have been added use `mctp link`
+commands. These will be 1:1 with the MCTP networks on the system.
+
+These objects host the interface `au.com.codeconstruct.MCTP.Network1`.
+
+### MCTP network interface: `au.com.codeconstruct.MCTP.Network1`
+
+All MCTP networks objects host the `au.com.codeconstruct.MCTP.Network1` dbus
+interface:
+
+```
+NAME                                 TYPE      SIGNATURE RESULT/VALUE FLAGS
+au.com.codeconstruct.MCTP.Network1  interface -         -            -
+.LocalEIDs                          property  ay        1 8          const
+```
+
+The D-Bus interface includes the `LocalEIDs` property which reports BMC local EIDs
+in the network.
+
 ## Endpoint objects: `/au/com/codeconstruct/networks/<net>/endpoints/<eid>`
 
 These objects represent MCTP endpoints that `mctpd` has either discovered
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -51,6 +51,7 @@
 #define MCTP_DBUS_IFACE_ENDPOINT "xyz.openbmc_project.MCTP.Endpoint"
 #define OPENBMC_IFACE_COMMON_UUID "xyz.openbmc_project.Common.UUID"
 #define CC_MCTP_DBUS_IFACE_INTERFACE "au.com.codeconstruct.MCTP.Interface1"
+#define CC_MCTP_DBUS_NETWORK_INTERFACE "au.com.codeconstruct.MCTP.Network1"
 
 // an arbitrary constant for use with sd_id128_get_machine_app_specific()
 static const char* mctpd_appid = "67369c05-4b97-4b7e-be72-65cfd8639f10";
@@ -286,6 +287,31 @@ static peer * find_peer_by_addr(ctx *ctx, mctp_eid_t eid, int net)
 	if (eid != 0 && n && n->peeridx[eid] >= 0)
 		return &ctx->peers[n->peeridx[eid]];
 	return NULL;
+}
+
+static int find_local_eids_by_net(ctx *ctx, uint32_t net,
+                                  size_t* local_eid_cnt,
+                                  mctp_eid_t *ret_eids)
+{
+	size_t local_count = 0;
+	net_det *n = lookup_net(ctx, net);
+	peer *peer;
+
+	*local_eid_cnt = 0;
+	if (!n)
+		return -EINVAL;
+
+	for (size_t t = 0; t < 256; t++) {
+		if (n->peeridx[t] < 0)
+			continue;
+
+		peer = &ctx->peers[n->peeridx[t]];
+		if (peer && (peer->state == LOCAL))
+			ret_eids[local_count++] = t;
+	}
+	*local_eid_cnt = local_count;
+
+	return 0;
 }
 
 /* Returns a deferred free pointer */
@@ -2915,6 +2941,26 @@ static bool is_endpoint_path(const char *path)
 	return true;
 }
 
+static bool get_networkid_from_path(const char *path, uint32_t* netid)
+{
+	char *netstr = NULL;
+	int rc;
+
+	rc = sd_bus_path_decode_many(path,
+				     MCTP_DBUS_PATH "/networks/%",
+				     &netstr);
+
+	if (rc <= 0)
+		return false;
+
+	dfree(netstr);
+
+	if (parse_uint32(netstr, netid) < 0)
+		return false;
+
+	return true;
+}
+
 static bool is_interfaces_path(const char *path)
 {
 	char *intfName = NULL;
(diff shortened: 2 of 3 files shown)


amboar:
I'm not trying to block the change here, but I feel we might want a holistic treatment of `Figure 9` from DSP0236 v1.3.3). I feel that requires a bit more thought about how EID associations are represented. I don't think the proposal is wrong as such, but it is incomplete.

jk-ozlabs:
From some chatting on Discord, my suggested approach was that we have an EIDs property on the network object, where an application can always pick the first to acquire a suitable local EID that can route to the local stack. Since the host will accept any local destination EID, it will always be valid on any incoming interface.

(the extension of that concept is that we may as well assign the same local EID to all interfaces, but that's entirely optional)

We *may* have a scenario in future where an application may need to query a local EID that is bound to a specific interface, but I don't think that's what's happening here.

ThuBaNguyen:
> From some chatting on Discord, my suggested approach was that we have an EIDs property on the network object, where an application can always pick the first to acquire a suitable local EID that can route to the local stack. 

It seems I missed your idea in the Discord Chat. I thought we will add the `EIDs` property to `au.com.CodeConstruct.MCTP.Interface1` D-Bus interface of `interface` D-Bus object such as `/au/com/codeconstruct/mctp1/interfaces/mctpi2c3`

In case, we want to add the `EIDs` property to network object path such as `/au/com/codeconstruct/mctp1/networks/1`, I wonder which D-Bus interface should we use to add `EIDs` property?
```
root@mtmitchell-dcscm:~# busctl introspect au.com.codeconstruct.MCTP1 /au/com/codeconstruct/mctp1/networks/1
NAME                                TYPE      SIGNATURE RESULT/VALUE FLAGS
org.freedesktop.DBus.Introspectable interface -         -            -
.Introspect                         method    -         s            -
org.freedesktop.DBus.Peer           interface -         -            -
.GetMachineId                       method    -         s            -
.Ping                               method    -         -            -
org.freedesktop.DBus.Properties     interface -         -            -
.Get                                method    ss        v            -
.GetAll                             method    s         a{sv}        -
.Set                                method    ssv       -            -
.PropertiesChanged                  signal    sa{sv}as  -            -
```

> Since the host will accept any local destination EID, it will always be valid on any incoming interface.

I'm agree with you that when for the interfaces are belong to the same network, this will be true. Although, this will make the routing table in both BMC, the bridge between BMC and terminus (if have) and the terminus are more complicated.
Ampere team is still evaluating this solution. I will give more opinions in this solution later.

> 
> (the extension of that concept is that we may as well assign the same local EID to all interfaces, but that's entirely optional)

This option can only be applied when BMC is BO in all interfaces.

> 
> We _may_ have a scenario in future where an application may need to query a local EID that is bound to a specific interface, but I don't think that's what's happening here.

ThuBaNguyen:
> From some chatting on Discord, my suggested approach was that we have an EIDs property on the network object, where an application can always pick the first to acquire a suitable local EID that can route to the local stack. Since the host will accept any local destination EID, it will always be valid on any incoming interface.

What do you think about adding the D-Bus interface "au.com.CodeConstruct.MCTP.Network1" in network object path as below
```
 busctl introspect au.com.codeconstruct.MCTP1 /au/com/codeconstruct/mctp1/networks/1
NAME                                TYPE      SIGNATURE RESULT/VALUE FLAGS
org.freedesktop.DBus.Introspectable interface -         -            -
.Introspect                         method    -         s            -
org.freedesktop.DBus.Peer           interface -         -            -
.GetMachineId                       method    -         s            -
.Ping                               method    -         -            -
org.freedesktop.DBus.Properties     interface -         -            -
.Get                                method    ss        v            -
.GetAll                             method    s         a{sv}        -
.Set                                method    ssv       -            -
.PropertiesChanged                  signal    sa{sv}as  -            -
au.com.CodeConstruct.MCTP.Network1  interface -         -                                      -
.EIDs                               property  ay        2 8 12                                    const
```
Where:
1. `au.com.CodeConstruct.MCTP.Network1` is interface name. `1` is version.
2. There is one properties in the `network` D-Bus interfaces:
    a) `EIDs` will be BMC local EIDs in the network 1 as we discuss before.

jk-ozlabs:
Looks good, but maybe make it obvious that the `EIDs` are local to this endpoint? we could name it `LocalEIDs`, perhaps?

ThuBaNguyen:
> Looks good, but maybe make it obvious that the `EIDs` are local to this endpoint? we could name it `LocalEIDs`, perhaps?

Sure. `LocalEIDs` is more details property name.

jk-ozlabs:
ok, super. Also, just check the capitalisation on the interface name.

src/mctpd.c
```diff
@@ -288,6 +289,53 @@ static peer * find_peer_by_addr(ctx *ctx, mctp_eid_t eid, int net)
...
+		if (n->peeridx[t] < 0)
+			continue;
+
+		peer = &ctx->peers[n->peeridx[t]];
+		if (peer && (peer->state == LOCAL))
+			ret_eids[local_count++] = t;
+	}
+
+	return ret_eids;
+}
```
jk-ozlabs, line 336: You have essentially the same logic repeated here. Just return the count and array from the one function.

src/mctpd.c
```diff
@@ -3165,6 +3284,16 @@ static const sd_bus_vtable bus_endpoint_link_vtable[] = {
 	SD_BUS_VTABLE_END
 };
 
+static const sd_bus_vtable bus_network_vtable[] = {
+	SD_BUS_VTABLE_START(0),
+	SD_BUS_PROPERTY("LocalEids",
```
jk-ozlabs, line 3289: Apologies for not picking this one up earlier, but this should be `LocalEIDs` to match the existing convention:

 * the `EID` property on `xyz.openbmc_project.MCTP.Endpoint`
 * the capitalisation of abbreviations (`MCTP`) elsewhere in property/interface names

Given that "eid" is also a word, we want to make it recognisable as the abbreviation.

docs/mctpd.md
```diff
@@ -143,6 +143,27 @@ Like SetupEndpoint but will not assign EIDs, will only query endpoints for a
...
 endpoint, or `true` when an endpoint's EID is newly discovered.
 
+## Network objects: `/au/com/codeconstruct/networks/<net>`
+
+These objects represent MCTP networks which have been added use `mctp link`
+commands. These will be 1:1 with the MCTP networks on the system.
+
+These objects host the interface `au.com.codeconstruct.MCTP.Network1`.
+
+### MCTP interface interface: `au.com.codeconstruct.MCTP.Network1`
```
jk-ozlabs, line 153: this should be "MCTP network interface", no?

jk-ozlabs (approved):
Looks good. Just a wording change on the doc addition, but I can edit that as part of the merge (unless you happen to beat me to it!)

Thanks for the contribution!

ThuBaNguyen:
The PULL is WIP

jk-ozlabs:
OK, I'll wait on updates then!

jk-ozlabs:
OK, if you're updating, then a couple of build warnings present. You may already be working on these, but just in case not:

```
[20/24] Compiling C object test-mctpd.p/src_mctpd.c.o
../src/mctpd.c: In function ‘bus_network_get_prop’:
../src/mctpd.c:3047:22: warning: ‘eids’ may be used uninitialized [-Wmaybe-uninitialized]
 3047 |                 rc = find_local_eids_by_net(ctx, netid, &num, eids);
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/mctpd.c:3039:21: note: ‘eids’ was declared here
 3039 |         mctp_eid_t *eids;
      |                     ^~~~
[21/24] Compiling C object mctpd.p/src_mctpd.c.o
../src/mctpd.c: In function ‘bus_network_get_prop’:
../src/mctpd.c:3047:22: warning: ‘eids’ may be used uninitialized [-Wmaybe-uninitialized]
 3047 |                 rc = find_local_eids_by_net(ctx, netid, &num, eids);
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/mctpd.c:3039:21: note: ‘eids’ was declared here
 3039 |         mctp_eid_t *eids;
      |                     ^~~~
```

ThuBaNguyen:
The code is good now. I tested the code.

src/mctpd.c
jk-ozlabs: This has been re-introduced in the recent changes; does this iteration code need to appear twice?

ThuBaNguyen: Yes.

We need to know the `number of localEIDs` first, then allocate ret_eids with `number of localEIDs`.
In the commit https://github.com/CodeConstruct/mctp/commit/17adff2e1672ac29e32cb1ccbb6731910001e040, I tried to change combine `find_local_eids_count_by_net` and `find_local_eids_by_net` but `eids` is null event `num` is 1 in ` sd_bus_message_append_array(reply, 'y', eids, num);`. That why I back to use two functions.

ThuBaNguyen: Using two functions also makes the code is clearer.

jk-ozlabs: > Using two functions also makes the code is clearer.

I disagree. You *already* have divergence between the two implementations (code formatting), as well as a bug duplicated in both (`&ctx->peers[x]` can never be null)

The number of local EIDs is fairly well constrained; does this even need to be calculated in advance? just malloc the whole 256 bytes, and iterate once.

> but `eids` is null event `num` is 1 in `sd_bus_message_append_array(reply, 'y', eids, num);`

not sure I understand this, can you rephrase?

ThuBaNguyen: > The number of local EIDs is fairly well constrained; does this even need to be calculated in advance? just malloc the whole 256 bytes, and iterate once.
> 

I updated the latest code by malloc the localEIDs with 256 bytes as your suggest.
https://github.com/CodeConstruct/mctp/commit/edc3f5f0ff75521fb71aa4989fdc36351d631f69
Only need one functions and iterate once.

jk-ozlabs (approved):
Looks good, thanks!

jk-ozlabs:
Merged as https://github.com/CodeConstruct/mctp/commit/6470f9efd06168d790d05e1407720a59c54c9785 , along with a set of tests for the dbus interface.

I had fixed up the commit title (`LocaEIDs` -> `LocalEIDs`), so GH isn't autoclosing this.

Thanks for the contribution!

PR #57, opened 2024-12-19
mctp-bench: Add a utility tool to send/receive messages and display throughput

Timon-L:
A utility tool that implements a sender and receiver to benchmark mctp throughput. This is split into a sender command and a receive command in the same executable.

The receive command waits for incoming messages, notifies the user of first message received and then prints out measured throughput periodically.

Example usage:
	mctp-bench recv

The sender command sends messages continously, the size of which can be set by the user, defaulting to 8 bytes.

Example usage:
	mctp-bench send len 1000 eid 8

### DIFF
--- a/src/mctp-bench.c
+++ b/src/mctp-bench.c
@@ -1,0 +1,460 @@
+#define _XOPEN_SOURCE 700
+#include <bits/time.h>
+#include <err.h>
+#include <errno.h>
+#include <poll.h>
+#include <signal.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <sys/poll.h>
+#include <sys/socket.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "mctp.h"
+
+struct mctp_bench_send_args {
+        mctp_eid_t eid;
+        size_t len;
+        int net;
+};
+
+struct msg_header {
+        uint16_t magic;
+        uint32_t seq_no;
+};
+
+struct mctp_stats {
+        size_t total_received_len, curr_packet_len;
+        uint32_t prev_seq_no;
+        float elapsed_time;
+        unsigned long msgs_dropped, msg_count, invalid_payloads;
+};
+
+struct recv_ctx {
+        struct mctp_stats stats;
+        struct timespec start_time, current_time;
+        unsigned char *buf;
+        bool started_recv_flag;
+        int sd;
+};
+
+static const size_t MSG_HEADER_LEN = sizeof(struct msg_header);
+static const size_t MAX_LEN = 64 * 1024 - 1;
+static const uint32_t SEQ_START = UINT32_MAX - 5;
+static const uint16_t MAGIC_VAL = 0xbeca;
+static const int DEFAULT_NET = MCTP_NET_ANY;
+static const int DEFAULT_SECONDS_INTERVAL = 10;
+
+static float get_throughput(float total_len, float elapsed_time)
+{
+        return total_len / (elapsed_time * 1024);
+}
+
+static void print_stats(struct recv_ctx *recv_ctx)
+{
+        float throughput = get_throughput(recv_ctx->stats.total_received_len,
+                                          recv_ctx->stats.elapsed_time);
+        printf("Throughput: %.2f kB/s | Recevied: %lu msgs | "
+               "Dropped: %lu msgs | "
+               "Invalid: %lu msgs\n",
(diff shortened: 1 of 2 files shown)


src/mctp-bench.c
```diff
@@ -0,0 +1,413 @@
...
+               "Invalid: %lu msgs\n",
+               throughput, recv_ctx->stats.msg_count,
+               recv_ctx->stats.msgs_dropped, recv_ctx->stats.invalid_payloads);
+}
+
+static int get_elapsed_time(struct recv_ctx *recv_ctx)
+{
+        return (recv_ctx->current_time.tv_sec - recv_ctx->start_time.tv_sec) +
+               (recv_ctx->current_time.tv_nsec - recv_ctx->start_time.tv_nsec) /
+                   1e9;
```
mkj, line 73: Should this return a float? Currently it'll truncate to seconds.

src/mctp-bench.c
```diff
@@ -0,0 +1,413 @@
...
+
+struct msg_header {
+        uint16_t magic;
+        uint32_t seq_no;
+};
+
+struct mctp_stats {
+        size_t total_received_len, curr_packet_len;
+        uint32_t prev_seq_no;
+        double elapsed_time;
```
mkj, line 33: Maybe `float` for consistency with other floating point variables? (Either is fine, but having them different makes me wonder if there's some reason I'm missing)

src/mctp-bench.c
```diff
@@ -0,0 +1,413 @@
...
+static float get_throughput(float total_len, float elapsed_time)
+{
+        return total_len / (elapsed_time * 1024);
+}
+
+static void print_stats(struct recv_ctx *recv_ctx)
+{
+        float throughput =
+            get_throughput((float)recv_ctx->stats.total_received_len,
+                           (float)recv_ctx->stats.elapsed_time);
```
mkj, line 61: Don't need `(float)` here?

src/mctp-bench.c
```diff
@@ -0,0 +1,413 @@
...
+                } else
+                        timeout = -1;
+
+                rc = poll(&pollfd, 1, timeout);
+                if (rc < 0) {
+                        warn("recv: poll failed");
+                        break;
+                }
+
+                if (pollfd.revents & POLLIN) {
```
mkj, line 198: I think it should probably test `rc == 1` from `poll`? (Not certain what happens if `poll()` times out, whether it clears `revents` or not)

src/mctp-bench.c
```diff
@@ -0,0 +1,413 @@
...
+        uint32_t prev_seq_no;
+        double elapsed_time;
+        unsigned long msgs_dropped, msg_count, invalid_payloads;
+};
+
+struct recv_ctx {
+        struct mctp_stats stats;
+        struct timespec start_time, current_time;
+        unsigned char *buf;
+        volatile bool started_recv_flag;
```
mkj, line 41: Don't need `volatile` (nothing is modifying memory behind the compiler's back, which is the case where `volatile` is _sometimes_ right to use)

src/mctp-bench.c
```diff
@@ -0,0 +1,413 @@
...
+        errno = 0;
+        *out = strtoul(opt, &endptr, 0);
+        if (endptr == opt || errno == ERANGE) {
+                return -1;
+        }
+        return 0;
+}
+
+static int parse_net_and_eid(struct mctp_bench_send_args *send_args, char *opt)
+{
```
mkj, line 295: We should put these helpers into `mctp-utils.c` eventually (doesn't need to happen in this PR)

mkj (commented):
Looks good, thanks. Noted a few small things

src/mctp-bench.c
Timon-L: Changed function and calculation to use floating point number to get higher precision for elapsed time.

src/mctp-bench.c
Timon-L: I missed that, thanks for pointing it out. Applied suggested change for consistency.

src/mctp-bench.c
Timon-L: Removed unnecessary casting.

src/mctp-bench.c
Timon-L: Applied suggested change, now it won't attempt to receive and handle the message if rc is not 1.

Timon-L:
Thanks for the feedback, applied the suggested changes and added allocate tag function to check if SIOCMCTPALLOCTAG2 or SIOCMCTPALLOCTAG exists for the sender to use.

mkj (approved):

PR #65, opened 2025-02-18
mctp-bench: Adjust headers

kraj:
bits/time.h is not be included directly, instead use time.h since poll.h is included there is no need to include sys/poll.h

Fixes
../git/src/mctp-bench.c:2:10: fatal error: 'bits/time.h' file not found
    2 | #include <bits/time.h>
      |          ^~~~~~~~~~~~~


jk-ozlabs:
Thanks for the contribution!

PR #72, opened 2025-04-28
Use Vendor message types for mctp-bench, mctp-echo

mkj:
Previously PLDM message type (1) was being used, but it would be better to use a specific type that won't conflict with other uses.

This could still clash with other vendor types, pending future kernel support to distinguish those.

### DIFF
--- a/src/mctp-bench.c
+++ b/src/mctp-bench.c
@@ -15,6 +15,10 @@
 
 #include "mctp.h"
 
+// Code Construct allocation
+static const uint8_t VENDOR_TYPE_BENCH[3] = { 0xcc, 0xde, 0xf1 };
+static const uint8_t MCTP_TYPE_VENDOR_PCIE = 0x7e;
+
 struct mctp_bench_send_args {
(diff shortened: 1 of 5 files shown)


src/mctp-bench.c
```diff
@@ -116,6 +121,12 @@ static int handle_incoming_msg(struct recv_ctx *recv_ctx)
         }
 
         hdr = (struct msg_header *)recv_ctx->buf;
+        if (memcmp(hdr->vendor_prefix, VENDOR_TYPE_BENCH, sizeof(VENDOR_TYPE_BENCH)) != 0) {
```
jk-ozlabs, line 124: We should check that we have at least `sizeof(VENDOR_TYPE_BENCH)`.

But I guess we're not checking that we have a full header either; we should probably add that, before the initial `(struct msg_header *)` cast.

mkj: I've added a check.

jk-ozlabs:
LGTM, thanks!

PR #77, opened 2025-05-30
Fix valgrind reported leak

mkj:
These changes should have no effect on program execution, just make it easier to see leaks in valgrind


src/mctpd.c
```diff
@@ -3867,11 +3886,13 @@ int main(int argc, char **argv)
 
 	sd_bus_flush(ctx->bus);
 	sd_bus_close(ctx->bus);
+	sd_bus_unref(ctx->bus);
```
jk-ozlabs, line 3891: looks like we have a `sb_bus_flush_unref_close()` here

mkj: Nice, thanks

PR #78, opened 2025-06-02
Allow passing config as a fixture

khangng-ampere:
This MR adds the capability to write tests for `mctpd` as an endpoint. This caught an uninitialized read bug on link data.

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -196,7 +196,7 @@ struct ctx {
 	sd_bus *bus;
 
 	// Configuration
-	const char *config_filename;
+	char *config_filename;
 
 	mctp_nl *nl;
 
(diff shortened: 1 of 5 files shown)


jk-ozlabs:
I'll see if I can reproduce here; if that's also passing, maybe a CI update would be handy...

src/mctpd.c
```diff
@@ -3573,7 +3573,7 @@ static int add_interface(struct ctx *ctx, int ifindex)
 		return -ENOENT;
 	}
 
-	struct link *link = malloc(sizeof(*link));
+	struct link *link = calloc(sizeof(*link), 1);
```
mkj, line 3576: Thanks for catching that! This should be `calloc(1, sizeof(*link))` (no real difference, but some recent GCCs complain).

khangng-ampere:
Seems like ASAN caught some more bugs, I can reproduce with `meson setup build -Db_sanitize=address` just like how CI does. Let me see what it is and push a fix.

src/mctpd.c
khangng-ampere: Ah yes, I messed up the order. Acknowledged.

khangng-ampere:
CI passed now. There was a memory leak when we copied the config file path from CLI args.

PR #79, opened 2025-06-02
QOL changes for test runner

khangng-ampere:
This PR is a collection of small QOL changes to the test runner:

1. Make TAP output a tad more useful, with captured logs from `mctpd` subprocess.
2. Allow running without TAP by passing `--interactive` to `meson`.
3. Allow filtering tests (or passing arbitrary flags) to `pytest` via `--test-args`.

As a result, instead of

```
meson test -C obj
cd obj
dbus-run-session env DBUS_STARTER_BUS_TYPE=user pytest
cd ..
```

It can be now

```
meson test -C obj --interactive
```

and for filtering tests:

```
meson test -C obj --interactive --test-args="-k recov"
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -124,7 +124,7 @@ if tests
     # redirecting to a file. We can detect this if stdout is an terminal, and
     # disable TAP protocol.
     pytest = find_program('pytest')
-    script = '''export DBUS_STARTER_BUS_TYPE=user; if [ ! -t 1 ]; then PYTEST_FLAGS="--tap"; fi; @0@ $PYTEST_FLAGS'''.format(pytest.full_path())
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -140,7 +140,7 @@ if tests
 
     test('test-mctpd', dbus_run_session,
         depends: mctpd_test,
-        args: [ sh.full_path(), '-c', script ],
+        args: [ sh.full_path(), '-c', script, '--' ],
```
jk-ozlabs, line 143: so meson's `--test-args` seems fairly sparsely documented; those arguments are getting appended to this list, right? (hence the `--` ?)

khangng-ampere: Yes, that is correct

PR #85, opened 2025-07-02
MCTP Bridge Polling

faizana-nvidia:
**MCTP Bridge Endpoint poll**

Reference :  DSP0236 section 8.17.6 Reclaiming EIDs from hot-plug devices

```The bus owner/bridge can detect a removed device or devices by validating the EIDs that are presently allocated to endpoints that are directly on the bus and identifying which EIDs are missing. It can do this by attempting to access each endpoint that the bridge has listed in its routing table as being a device that is directly on the particular bus. Attempting to access each endpoint can be accomplished by issuing the Get Endpoint ID command.```

This PR aims to extend ```mctpd``` as bus owner to enhance bridge support feature by incorporating downstream endpoint polling.
One way to detect the accessibility of its downstream devices is for bus owner/bridge to probes on each bridge's downstream endpoint using the Get Endpoint ID command.

Once its established that endpoint path is accessible, we would stop polling to that eid and publish the eid endpoint object

This PR is based  discussions over MCTP Bridge Support : https://github.com/CodeConstruct/mctp/pull/71

### DIFF
--- a/tests/test_mctpd.py
+++ b/tests/test_mctpd.py
@@ -1506,3 +1506,205 @@ async def test_query_peer_properties_retry_timeout(nursery, dbus, sysnet):
     # exit mctpd
     res = await mctpd.stop_mctpd()
     assert res == 0
+""" Test that we use endpoint poll interval from the config and
+    that we discover bridged endpoints via polling"""
+async def test_bridged_endpoint_poll(dbus, sysnet, nursery, autojump_clock):
+    poll_interval = 2500
+    config = f"""
+    [bus-owner]
+    endpoint_poll_ms = {poll_interval}
+    """
+
+    mctpd = MctpdWrapper(dbus, sysnet, config = config)
+    await mctpd.start_mctpd(nursery)
+
+    iface = mctpd.system.interfaces[0]
+    ep = mctpd.network.endpoints[0]
+    mctp = await mctpd_mctp_iface_obj(dbus, iface)
+
+    bridged_ep = [
+        Endpoint(iface, bytes(), types = [0, 1]),
+        Endpoint(iface, bytes(), types = [0, 1])
+    ]
+    for bep in bridged_ep:
+        mctpd.network.add_endpoint(bep)
+        ep.add_bridged_ep(bep)
+
+    mctp_obj = await dbus.get_proxy_object(MCTPD_C, MCTPD_MCTP_P)
+    mctp_objmgr = await mctp_obj.get_interface(DBUS_OBJECT_MANAGER_I)
+    endpoint_added = trio.Semaphore(initial_value=0)
+
+    # We expect two bridged endpoints to be discovered
+    expected_bridged_eps = len(bridged_ep)
+    bridged_endpoints_found = []
+
+    def ep_added(ep_path, content):
+        if MCTPD_ENDPOINT_I in content:
+            bridged_endpoints_found.append(ep_path)
+            endpoint_added.release()
+
+    await mctp_objmgr.on_interfaces_added(ep_added)
+    (eid, _, path, new) = await mctp.call_assign_endpoint(ep.lladdr)
+    assert new
+
+    # Wait for all expected bridged endpoints to be discovered
+    with trio.move_on_after(poll_interval / 1000 * 2) as expected:
+        for i in range(expected_bridged_eps):
+            await endpoint_added.acquire()
+
+    # Verify we found all expected bridged endpoints
+    assert not expected.cancelled_caught, "Timeout waiting for bridged endpoints"
+    assert len(bridged_endpoints_found) == expected_bridged_eps
+
+    res = await mctpd.stop_mctpd()
+    assert res == 0
+
+""" Test that all downstream endpoints are removed when the bridge
+    endpoint is removed"""
+async def test_bridged_endpoint_remove(dbus, sysnet, nursery, autojump_clock):
+    poll_interval = 2500
+    config = f"""
+    [bus-owner]
+    endpoint_poll_ms = {poll_interval}
+    """
+
+    mctpd = MctpdWrapper(dbus, sysnet, config = config)
+    await mctpd.start_mctpd(nursery)
+
+    iface = mctpd.system.interfaces[0]
+    ep = mctpd.network.endpoints[0]
+    mctp = await mctpd_mctp_iface_obj(dbus, iface)
+
+    bridged_ep = [
+        Endpoint(iface, bytes(), types = [0, 1]),
+        Endpoint(iface, bytes(), types = [0, 1])
+    ]
+    for bep in bridged_ep:
+        mctpd.network.add_endpoint(bep)
+        ep.add_bridged_ep(bep)
+
+    (eid, _, path, new) = await mctp.call_assign_endpoint(ep.lladdr)
+    assert new
+
+    # Wait for the bridged endpoints to be discovered
+    await trio.sleep((poll_interval * 2) / 1000)
+    removed = trio.Semaphore(initial_value = 0)
+    removed_eps = []
+
+    # Capture the removed endpoints
+    def ep_removed(ep_path, interfaces):
+        if MCTPD_ENDPOINT_I in interfaces:
+            removed.release()
+            removed_eps.append(ep_path)
+
+    mctp_obj = await dbus.get_proxy_object(MCTPD_C, MCTPD_MCTP_P)
+    mctp_objmgr = await mctp_obj.get_interface(DBUS_OBJECT_MANAGER_I)
+    await mctp_objmgr.on_interfaces_removed(ep_removed)
+
+    # Remove the bridge endpoint
+    bridge_obj = await mctpd_mctp_endpoint_control_obj(dbus, path)
+    await bridge_obj.call_remove()
+
+    # Assert that all downstream endpoints were removed
+    assert len(removed_eps) == (len(bridged_ep) + 1)
+    res = await mctpd.stop_mctpd()
+    assert res == 0
+
+""" Test that polling stops once endponit has been discovered """
+async def test_bridged_endpoint_poll_stop(dbus, sysnet, nursery, autojump_clock):
+    poll_interval = 2500
+    config = f"""
+    [bus-owner]
+    endpoint_poll_ms = {poll_interval}
+    """
+
+    mctpd = MctpdWrapper(dbus, sysnet, config = config)
+    await mctpd.start_mctpd(nursery)
+
+    iface = mctpd.system.interfaces[0]
+    ep = mctpd.network.endpoints[0]
+    mctp = await mctpd_mctp_iface_obj(dbus, iface)
+    poll_count = 0
+
+    class BridgedEndpoint(Endpoint):
+        async def handle_mctp_control(self, sock, src_addr, msg):
+            flags, opcode = msg[0:2]
+            if opcode == 0x2: # Get Endpoint ID
+                nonlocal poll_count
+                poll_count += 1
+            return await super().handle_mctp_control(sock, src_addr, msg)
+
+    bridged_ep = BridgedEndpoint(iface, bytes(), types = [0, 1])
+    mctpd.network.add_endpoint(bridged_ep)
+    ep.add_bridged_ep(bridged_ep)
+
+    (eid, _, path, new) = await mctp.call_assign_endpoint(ep.lladdr)
+    assert new
+
+    mctp_obj = await dbus.get_proxy_object(MCTPD_C, MCTPD_MCTP_P)
+    mctp_objmgr = await mctp_obj.get_interface(DBUS_OBJECT_MANAGER_I)
+    endpoint_added = trio.Semaphore(initial_value=0)
+    poll_count_by_discovery = 0
+
+    def ep_added(ep_path, content):
+        if MCTPD_ENDPOINT_I in content:
+            nonlocal poll_count_by_discovery
+            poll_count_by_discovery = poll_count
+            endpoint_added.release()
+
+    await mctp_objmgr.on_interfaces_added(ep_added)
+
+    # Wait longer than the poll interval for the bridged endpoint
+    # to be discovered
+    await trio.sleep(poll_interval / 1000)
+
+    # We should have only poll until the discovery thus count should
+    # be the same even after longer wait.
+    assert poll_count == poll_count_by_discovery
+
+    res = await mctpd.stop_mctpd()
+    assert res == 0
+
+""" Test that polling continues until the endpoint is discovered """
+async def test_bridged_endpoint_poll_continue(dbus, sysnet, nursery, autojump_clock):
+    poll_interval = 2500
+    config = f"""
+    [bus-owner]
+    endpoint_poll_ms = {poll_interval}
+    """
+
+    mctpd = MctpdWrapper(dbus, sysnet, config = config)
+    await mctpd.start_mctpd(nursery)
+
+    iface = mctpd.system.interfaces[0]
+    ep = mctpd.network.endpoints[0]
+    mctp = await mctpd_mctp_iface_obj(dbus, iface)
+    poll_count = 0
+
+    class BridgedEndpoint(Endpoint):
+        async def handle_mctp_control(self, sock, src_addr, msg):
+            flags, opcode = msg[0:2]
+            # dont respond to simiulate device not accessible
+            # but increment poll count for the Get Endpoint ID
+            if opcode == 0x2: # Get Endpoint ID
+                nonlocal poll_count
+                poll_count += 1
+            return None
+
+    bridged_ep = BridgedEndpoint(iface, bytes(), types = [0, 1])
+    mctpd.network.add_endpoint(bridged_ep)
+    ep.add_bridged_ep(bridged_ep)
+
+    (eid, _, path, new) = await mctp.call_assign_endpoint(ep.lladdr)
+    assert new
+
+    # Wait for sometime to continue polling
+    await trio.sleep(poll_interval / 1000)
+
+    poll_count_before = poll_count
+    # Wait more to see if poll count increments
+    await trio.sleep(1)
+    assert poll_count > poll_count_before
+
+    res = await mctpd.stop_mctpd()
+    assert res == 0

### CODE REVIEW

**Comment by faizana-nvidia** [comment]
I also have added a TODO comment here [todo](https://github.com/CodeConstruct/mctp/pull/85/commits/955d1b4c08fe89b2cc02620820125ce4e0a6b5ce#diff-222eb3ffb54581db6e3d72a6efa39bf25a8145a7f8eab9b7f4474ea184675da9R3790)

in our current implementation of Busowner pool and local eid assignment, we don't have any check to see, if localEID which is being set via mctp tool **(mctp addr add \<eid> dev \<interface>)** for any network, has already been used by Busowner when it assigned dynamic/static EID to any MCTP device/endpoint. This makes current behavior mess up the network peers[] structure as overall Busowner EID space is common to local and endpoint EIDs. I have been meaning to raise an issue to ask this question.

Since we are getting into the reservation of EIDs, our new localEID could very well be from the reserved EID space. So should we introduce restriction of such manner in mctp tool?

**Comment by jk-ozlabs** [comment]
> May be we can introduce a new var representing actual_pool_size (response of SET_ENDPOINT ID) and call out these two vars, peer->pool_start and peer->pool_size as allocated data while network reserved_eid_set continues to interact with these vars as it has been doing.

Yes, that's along the lines of what I was thinking. But maybe with the names adjusted: `pool_start` & `pool_size` would represent actually-allocated values, `req_pool_size` and `req_pool_start` for the (temporarily-valid) request details.

And then as an improvement: since the `req_` values are transient, is there any better way to pass them from the Set Endpoint ID response handler to the Allocate Endpoint IDs command? (ie., rather than setting transient values on the struct peer)

> I also have added a TODO comment here 

There's not much we can do about that case; mctpd isn't (currently) managing that local EID, so there is not much we can do on conflicts - as you have noted.

> So should we introduce restriction of such manner in mctp tool/mctpd?

No, not while mctpd is not responsible for assigning local EIDs.

I see the best way to prevent this case is a facility to explicitly configure mctpd with the range of dynamically-allocatable EIDs, and it's up to the platform integrator to ensure that that range does not conflict with the potential set of local EIDs.

(that's one of the reasons for the configuration file support, for exactly this use-case)

**Comment by faizana-nvidia** [comment]
> And then as an improvement: since the `req_` values are transient, is there any better way to pass them from the Set Endpoint ID response handler to the Allocate Endpoint IDs command? (ie., rather than setting transient values on the struct peer)

1) To add further onto this, we don't know how many birdges on what bindings would turn up on network. Predicting their numbers would rather be tedious, so would be passing each pool size to their respective Allocate EndpointID. So in worse case scenario assuming 255 bridges per network(similar to peer numbers) we can have a separate array structure representing bridge data/pool_size which would be filled right after SET_ENDPOINT_ID.


2) One another way (though not tested) is somehow find the capacity of bridge just before asking ALLOCATE_ENDPOINT_ID. Unfortunately we need to know this because we want to avoid asking endpoints number in request which would be more that Bridge's dynamic pool size.

Table 23 – Allocate Endpoint IDs message

_An error completion code (ERROR_INVALID_DATA should be returned) shall be returned if the number of EIDs being allocated (Number of Endpoint IDs) exceeds the Dynamic Endpoint ID Pool size. (This error condition does not apply to when the number of endpoint IDs passed in the request is 0x00)._ 


3) We can also think this response of pool_size in SET_EDNPOINT_ID (transient) to be the capacity of bridge while the allocated pool size and start representing allocated values

> > I also have added a TODO comment here
> 
> There's not much we can do about that case; mctpd isn't (currently) managing that local EID, so there is not much we can do on conflicts - as you have noted.
> 
> > So should we introduce restriction of such manner in mctp tool/mctpd?
> 
> No, not while mctpd is not responsible for assigning local EIDs.
> 
> I see the best way to prevent this case is a facility to explicitly configure mctpd with the range of dynamically-allocatable EIDs, and it's up to the platform integrator to ensure that that range does not conflict with the potential set of local EIDs.
> 
> (that's one of the reasons for the configuration file support, for exactly this use-case)

Acknowledged.

**Comment by jk-ozlabs** [comment]
> we can have a separate array structure representing bridge data/pool_size which would be filled right after SET_ENDPOINT_ID.

If there is no way to better handle the transient requested peer size, I would prefer to just store it on the `struct peer`.

> One another way (though not tested) is somehow find the capacity of bridge just before asking ALLOCATE_ENDPOINT_ID

We have the capacity of the brige, from the Set Endpoint ID response, no? As you say:

> We can also think this response of pool_size in SET_EDNPOINT_ID (transient) to be the capacity of bridge while the allocated pool size and start representing allocated values

I would suggest:

 1. record the requested pool size from the Set Endpoint ID response
 2. attempt to allocate that requested pool size (pending any configured limits), by querying the `struct peer` array
 3. record the allocated pool range on that peer
 4. perform Allocate Endpoint IDs from that allocated pool

If the allocation is not possible in (2), just fail for now. We can look at fallbacks in a follow-up.

I think this it fine to do without a new dbus call, but am open to ideas about why we may need one.

**Comment by faizana-nvidia** [comment]
> > we can have a separate array structure representing bridge data/pool_size which would be filled right after SET_ENDPOINT_ID.
> 
> If there is no way to better handle the transient requested peer size, I would prefer to just store it on the `struct peer`.
> 
> > One another way (though not tested) is somehow find the capacity of bridge just before asking ALLOCATE_ENDPOINT_ID
> 
> We have the capacity of the brige, from the Set Endpoint ID response, no? As you say:
> 
> > We can also think this response of pool_size in SET_EDNPOINT_ID (transient) to be the capacity of bridge while the allocated pool size and start representing allocated values
> 
> I would suggest:
> 
> 1. record the requested pool size from the Set Endpoint ID response
> 2. attempt to allocate that requested pool size (pending any configured limits), by querying the `struct peer` array
> 3. record the allocated pool range on that peer
> 4. perform Allocate Endpoint IDs from that allocated pool
> 
> If the allocation is not possible in (2), just fail for now. We can look at fallbacks in a follow-up.
> 
> I think this it fine to do without a new dbus call, but am open to ideas about why we may need one.

Please let me know if I understood you correctly on steps
1. Fetch out the transient pool size (via SET_ENDPOINT_ID).
2. Try to allocate (find the pool sized chunked of peer structure which are available to be later used by downstream EID via get_pool_start())
3. If found then update the peer structure with req_pool_size = capacity and req_pool_start = get_pool_start()
4. Also mark all such eids to be reserved
5. From dbus api AssignBridgeStatic get pool_size and try to allocate that size with req_pool_start and whatever is coming as response from ALLOCATE_EID command update peer structure with alloc_pool_size = response(pool_size) and alloc_pool_start=response(pool_start).

Introduces req_pool_size and req_pool_start, updates pool_size -> alloc_pool_size and pool_start ->alloc_pool_start.

If we are going by this then do we have to get pool_start argument from dbus method AssignBridgeStatic if we have already decided what pool start will be from req_pool_start?

Or do you mean not to use any new dbus method AssignBridgeStatic at all?
> I think this it fine to do without a new dbus call, but am open to ideas about why we may need one.

**Comment by jk-ozlabs** [comment]
Since the allocation mechanism is now in #71, I'll shift this discussion to there.

**Comment by faizana-nvidia** [comment]
Hello Jeremy,
I've rebased the PR and updated the logic based on other PR [71](https://github.com/CodeConstruct/mctp/pull/71) comments. Please help me with the review.

**Comment by jk-ozlabs** [comment]
I'm going to hold off on reviewing this one until we have the core bridge support done, given that's a hard dependency for this.

**Comment by faizana-nvidia** [comment]
> I'm going to hold off on reviewing this one until we have the core bridge support done, given that's a hard dependency for this.

understood, sounds good

**Comment by faizana-nvidia** [comment]
Hello Jeremy,
I have some questions on discussed approach of Polling [(discussion)](https://github.com/CodeConstruct/mctp/pull/71#issuecomment-2933806456) mechanism for MCTP Bridge downstream endpoint management.
To give you a better picture on MCTP Network, we have MCU as MCTP Bridge(over USB) managing GPUs at its downstream 
```mermaid
graph LR
    A["HMC/BMC"] --> B["MCTP Bridge"]
    B --> C["GPU1"]
    B --> D["GPU2"]                                          
```
I'm aware that in the [(discussed)](https://github.com/CodeConstruct/mctp/pull/71#issuecomment-2933806456) polling approach, mctpd doesn't care about state of devices behind the bridge once they have been discovered.
In our platform many of our applications such as Nvidia NSM (used for GPU Telemetry), relies directly on Event Subscription model or something similar which creates a requirement to explicitly do a re-subscribe to endpoints again when they undergoing reset. Owing to this and among other similar reasons, this current PR ([MCTP Bridge Polling](https://github.com/CodeConstruct/mctp/pull/85) )proposed for keeping the polling active throughout the existence of MCTP Bridge itself and inform to other d-bus endpoint object monitoring applications about case where MCTP device has fallen off the network. This would greatly help these applications to initiate their initial discovery sequence as early/quickly as possible without losing out valuable telemetry data since they would be made aware of such reset via MCTPD.

If this seems like a legit reason to you for which we can think of keeping polling, I have further question on implementation

The particular use case which might not get solved even with Polling is fast reset of downstream endpoints such as GPUs themselves which happens in time-window of couple of milliseconds (100-200 msec) since polling window is not be narrow enough to capture such fast reset and from MCTP perspective endpoint never gotten off the MCTP network. In such case NSM would not even get notified about this fast-reset of GPU.

Need some help with figuring out how do we expect endpoint to notify or BusOwner to become aware of such fast reset use case.

I'm open to any further guidance/opinion on Polling mechanism itself, or should MCTP even care about this etc. Or may be help me understand this better on the philosophy of mctpd not caring about state of device downstream to bridge or direct-peer itself.

faizana-nvidia:
Hello Jeremy,

Updated the PR with last discussion on stopping of polling. It's up for review now.

Sorry to be reluctant on this, but is there any plan to support Discovery Notify for direct peer to BMC as bus owner specifically for something like USB binding where we don't explicitly have methods like I2C bus detection to identify endpoint presence?

jk-ozlabs:
Thanks for the updates.

You now seem to have a bunch of fixes/changes to the bridging implementation, which don't belong in the polling feature addition.

Can you separate those into a different PR please?

src/mctpd.c
```diff
@@ -1862,6 +1907,13 @@ static int remove_peer(struct peer *peer)
 		sd_event_source_unref(peer->recovery.source);
 	}
 
+	if (peer->pool_size) {
```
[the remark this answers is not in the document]
faizana-nvidia: ACK, I should remove pctx too here

src/mctpd.c
```diff
@@ -4763,6 +4773,139 @@ static int endpoint_send_allocate_endpoint_ids(
...
+static int bridge_poll_start(struct peer *bridge)
+{
+	struct ctx *ctx;
+	int rc;
+	mctp_eid_t pool_start = bridge->pool_start;
+	mctp_eid_t pool_size = bridge->pool_size;
+	int i;
+	ctx = bridge->ctx;
+	sd_event_source **sources =
+		calloc(pool_size, sizeof(sd_event_source *));
```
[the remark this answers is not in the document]
faizana-nvidia: understood, will update

src/mctpd.c
```diff
@@ -4763,6 +4773,139 @@ static int endpoint_send_allocate_endpoint_ids(
...
+
+	 2. If endpoint is not present or doesn't responds to send poll
+	    commmand, then it has not been establed yet that endpoint
+		path from the bridge is accessible or not, thus continue
+		to poll.	
+	 */
+
+	if (!bridge) {
+		warnx("Bridge (eid %d) removed?", bridge->eid);
+		goto exit;
```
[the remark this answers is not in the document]
faizana-nvidia: okay sure

faizana-nvidia:
Sorry for  the delay in addressing the review comments. I'll push the changes soon, meanwhile could you help me with my responses

faizana-nvidia (commented):
I found one problem here, looks like kernel is not deleting the routes by itself for gateway routes. And we would avoid deleting routes if have_route flag is not set.

Should we consider removing from mctpd for now?

> static int del_interface(struct link *link)
{
	struct ctx *ctx = link->ctx;
	int ifindex = link->ifindex;
	if (ctx->verbose) {
		fprintf(stderr, "Deleting interface #%d\n", ifindex);
	}
	for (size_t i = 0; i < ctx->num_peers;) {
		struct peer *p = ctx->peers[i];
		if (p->state == REMOTE && p->phys.ifindex == ifindex) {
			int rc;
			// Linux removes routes to deleted links, so no need
			// to request removal.
			p->have_neigh = false;
			p->have_route = false;

jk-ozlabs:
> Is there a way to increase the time frame for unit tests beyond 30sec

Possibly, but I think the better approach here would be to decouple from wall-clock time in the test framework. There has been some work on that recently, but I don't think we're at a point where that's solved.

I'll take a look; if that's going to be longer-term then we may just want to temporarily increase the timeout...

faizana-nvidia:
> > Is there a way to increase the time frame for unit tests beyond 30sec
> 
> Possibly, but I think the better approach here would be to decouple from wall-clock time in the test framework. There has been some work on that recently, but I don't think we're at a point where that's solved.
> 
> I'll take a look; if that's going to be longer-term then we may just want to temporarily increase the timeout...

Hello @jk-ozlabs 

Gentle reminder, Will this block us for re-review this PR?

jk-ozlabs:
OK, the timers support has been merged. Let me know how you go with that.

We may want to discuss interactions with Get Routing Table Entries too.

faizana-nvidia:
> OK, the timers support has been merged. Let me know how you go with that.
> 

It worked like a charm :), test cases are running pretty fast. PR has been updated
> We may want to discuss interactions with Get Routing Table Entries too.

I see the PR [128](https://github.com/CodeConstruct/mctp/pull/128) opened for GRTE command support, will go though it

Pulling in your conversation here https://github.com/CodeConstruct/mctp/pull/128#issuecomment-3471123399

> _Perhaps we could use the GRTE responses as first-pass data on what to start polling, plus a configuration for "just poll the entire downstream pool" for cases where bridges do not support (or do not reliably support) GRTE_

For interacting with GRTE, as you mentioned in your above comment, idea is, to reply on polling only when GRTE isn't supported by Bridge.

Question :

If GRTE Supported - should polling continue only for eids (among the allocated pool space)which are not present as routing entry of bridge's routing table? if so then what happens for the case  where some endpoint (among the pool range)starts responding to polling commands, will we ask bridge to update its routes (may be using Routing Information Update)

If not Supported - continue as this PR is doing.

jk-ozlabs:
Yep, I think that's good summary of the logic there :)

Since it's an optimisation change (reducing polling requirements) rather than correctness (we'll still publish the saame enumeration results), I am OK with separating the GRTE changes until after the polling implementation, if you prefer.

src/mctpd.c
```diff
@@ -4752,6 +4756,26 @@ static int parse_config_bus_owner(struct ctx *ctx, toml_table_t *bus_owner)
 			return rc;
 	}
 
+	val = toml_int_in(bus_owner, "endpoint_poll_ms");
+	if (val.ok) {
+		uint64_t i = val.u.i;
+		if (i == 0) {
+			ctx->endpoint_poll = 0;
+			fprintf(stderr, "Bridge poll is disabled\n");
```
jk-ozlabs: Super minor, but we probably don't need this output by default. You could make it conditional on `ctx->verbose`, or remove it altogether.

faizana-nvidia:
> I think the structure here is sensible, just a few queries and changes inline.

Thanks for the review comments, I'll address to them shortly

src/mctpd.c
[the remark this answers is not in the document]
faizana-nvidia: I'll squash the commits

src/mctpd.c
```diff
@@ -2355,6 +2355,52 @@ static int get_endpoint_peer(struct ctx *ctx, sd_bus_error *berr,
...
+ * It can do this by attempting to access each endpoint that the bridge
+ * has listed in its routing table as being a device that is directly on
+ * the particular bus. Attempting to access each endpoint can be accomplished
+ * by issuing the Get Endpoint ID command...
+
+
+ * since bridged endpoints are routed from bridge, direct query
+ * to eid should work if gateway routes are in place.
+ */
+static int query_endpoint_poll_commmand(struct peer *peer, mctp_eid_t *resp_eid)
```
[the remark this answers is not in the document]
faizana-nvidia: right, let me merge the changes

src/mctpd.c
faizana-nvidia: Ack,  l'll make it conditional

tests/mctpenv/__init__.py
```diff
@@ -259,11 +259,21 @@ def find_endpoint(self, addr):
...
-
-            neigh = self.lookup_neighbour(route.iface, addr.eid)
-            # if no neighbour, return an empty lladdr (eg mctpusb)
-            lladdr = neigh.lladdr if neigh else bytes()
+            if route.gw is not None:
+                # In case of gateway routes, we need not have neighbours
+                # for the gated endpoints, but only need to find the
+                # gateway's physical address
+                gw_net, gw_eid = route.gw
+                gw_route = self.lookup_route(gw_net, gw_eid)
```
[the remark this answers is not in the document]
faizana-nvidia: Although for the above case in test framework, we would need to find a recursive lookup among Endpoint's bridged_eid so pass on incoming packet to expected downstream eid 

This is for forwarding the packets:  first based on routes find the interface and eid this needs to go toand then to right Bridge eid which will further forward to the right endpoint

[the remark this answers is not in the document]
jk-ozlabs: > Wondering if we would really see this recursive gateway route case.

It's certainly the most common case to have the one gateway route handling the entire pool. However, the kernel route implemenation does allow for nesting, to allow us to explicitly describe the full routing topology. For example, this may be useful to describe cases where a bridge (and its subordinate devices) can be routable over different paths. In that case, we can describe that alternate route - possibly with a higher priority - exactly as the hardware connections work.

We don't set up routes like that in mctpd currently though, so perhaps support for that would be OK to defer. But if we do that, we should indicate with a TODO comment.

conf/mctpd.conf
```diff
@@ -16,3 +16,6 @@ message_timeout_ms = 30
 dynamic_eid_range = [8, 254]
 
 max_pool_size = 15
+
+# Bus Owner/bridge polling interval (ms) to check endpoint accessibility.
+endpoint_poll_ms = 0 # disable poll
```
jk-ozlabs: Super minor, but maybe document the 0 = disable logic in the top-level comment, so that it isn't lost when someone actually configures the value. Perhaps:

```
# Bus Owner/bridge polling interval (in milliseconds) to check endpoint acccessibility
# A value of 0 will disable polling
endpoint_poll_ms = 0
```

src/mctpd.c
```diff
@@ -5213,6 +5328,139 @@ static int endpoint_send_allocate_endpoint_ids(
...
+	peer = n->peers[ep];
+	if (!peer) {
+		rc = add_peer(bridge->ctx, &(bridge->phys), ep, bridge->net,
+			      &peer, true);
+		if (rc < 0)
+			goto exit;
+	}
+
+	rc = query_endpoint_poll_commmand(peer, &ret_eid);
+	if (rc < 0) {
```
jk-ozlabs: Should we reschedule on all failures, or just timeouts?

src/mctpd.c
```diff
@@ -1963,6 +1974,57 @@ static int check_peer_struct(const struct peer *peer, const struct net *n)
...
+	pool_end = bridge->pool_start + bridge->pool_size - 1;
+	n = lookup_net(bridge->ctx, bridge->net);
+	pool_start = bridge->pool_start;
+	for (ep = pool_start; ep <= pool_end; ep++) {
+		// stop endpoint polling before removing peer
+		// else next trigger will create peer again.
+		int idx = ep - pool_start;
+
+		if (bridge->bridge_ep_poll.sources &&
+		    bridge->bridge_ep_poll.sources[idx]) {
```
jk-ozlabs: You have many uses of `bridge->bridge_ep_poll.sources[idx]` here, use a temporary.

... and no need to repeat the check for `bridge->bridge_ep_poll.sources` on every iteration, just do it once.

CHANGELOG.md
```diff
@@ -59,6 +59,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 9. `mctpd`'s Network1.LearnEndpoint method will now check for endpoint
    presence before publishing an endpoint.
 
+10. `mctpd` now implements bridge endpoint polling after successful dynamic pool
```
jk-ozlabs: This appears to have been added to the previous release, we want it in the top section.

For that reason: If you prefer, I am happy to manage the CHANGELOG entries. But if you leave this commit here, there needs to be a space after the `:` in the commit title.

jk-ozlabs (changes requested):
Mostly minor things there, but we do need to clarify the addition of an peer (through `add_peer`) before we know that the peer actually exists. We may not need the struct peer during polling at all.

Also, a couple of typos in the commit log: `endpoin` and `repond`. I don't have a facility to comment on the changelogs themselves, sorry!

src/mctpd.c
faizana-nvidia: Ack

src/mctpd.c
```diff
@@ -1963,6 +1974,57 @@ static int check_peer_struct(const struct peer *peer, const struct net *n)
 	return 0;
 }
 
+/* Stops downstream endpoint polling and removes
+ * peer structure when bridge endpoint is being removed.
+ */
+static int remove_bridged_peers(struct peer *bridge)
```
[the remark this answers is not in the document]
faizana-nvidia: its generic method to purge out bridged endpoint when bridge is going away, I should rather remove comments to avoid the confusion

CHANGELOG.md
faizana-nvidia: Thats fine, I'll just remove this commit, I feel you could manage the release info better.

src/mctpd.c
faizana-nvidia: gentle reminder for this query @jk-ozlabs

src/mctpd.c
```diff
@@ -5213,6 +5328,139 @@ static int endpoint_send_allocate_endpoint_ids(
...
+	 2. If endpoint is not present or doesn't responds to send poll
+	    commmand, then it has not been establed yet that endpoint
+		path from the bridge is accessible or not, thus continue
+		to poll.
+	 */
+
+	n = lookup_net(bridge->ctx, bridge->net);
+	peer = n->peers[ep];
+	if (!peer) {
+		rc = add_peer(bridge->ctx, &(bridge->phys), ep, bridge->net,
```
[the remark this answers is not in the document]
jk-ozlabs: > we don't have much option then, need to utilise raw method _endpoint_query_addr()

Yes, I think that may be best.

however, not sure I understand:

> may be create a new poll method which does entire mctp structure population then sends it over,

What "entire mctp structure" is this referring to? We can use the poll method that you have defined here (`peer_endpoint_poll`), but using `endpoint_query_addr` as the actual transfer mechanism. No?

src/mctpd.c
jk-ozlabs: Isn't that the opposite of what I had suggested?

I would be okay with that, with sufficient reasoning. Why should we not reschedule on other errors?

faizana-nvidia: > Should we reschedule on all failures, or just timeouts?

Sorry I got the impression we don't want reschedule for all errors (which I had in my previous version).
Reason why I felt this being okay was, like you stated if for unsuccessful case (for whatever reason) get endpoint id fails, likely next consecutive commands will also fail.

Im okay to update it to reschedule for other error cases as well owing to possibility of recovering in cases, such as :
some mismatch response which might have come in between 
or some send failure in the socket etc

faizana-nvidia: updated to reschedule for all errors

faizana-nvidia: yes, that's kept intentional, lemme know if you see any concern there.

jk-ozlabs:
Merged, thanks!

faizana-nvidia:
> Merged, thanks!

Thank you for all the help with code reviews and design discussions  :)

PR #91, opened 2025-07-12
mctpd: Fix IID of MCTP Control response message

gtossk2:
Fix incorrect IID of an MCTP Response message and it should match to the IID of an MCTP Request message.

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -64,7 +64,7 @@ static mctp_eid_t eid_alloc_max = 0xfe;
 static size_t MAX_PEER_SIZE = 1000000;
 
 static const uint8_t RQDI_REQ = 1 << 7;
-static const uint8_t RQDI_RESP = 0x0;
+static const uint8_t RQDI_RESP_MASK = 0x7f;
 static const uint8_t RQDI_IID_MASK = 0x1f;
 
 struct dest_phys {
@@ -645,7 +645,7 @@ static int handle_control_set_endpoint_id(struct ctx *ctx, int sd,
 	req = (void *)buf;
 
 	resp->ctrl_hdr.command_code = req->ctrl_hdr.command_code;
-	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP;
+	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP_MASK & req->ctrl_hdr.rq_dgram_inst;
 	resp->completion_code = 0;
 	resp->status = 0x01 << 4; // Already assigned, TODO
 	resp->eid_set = local_addr(ctx, addr->smctp_ifindex);
@@ -696,7 +696,7 @@ handle_control_get_version_support(struct ctx *ctx, int sd,
 	}
 
 	resp->ctrl_hdr.command_code = req->ctrl_hdr.command_code;
-	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP;
+	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP_MASK & req->ctrl_hdr.rq_dgram_inst;
 	return reply_message(ctx, sd, resp, resp_len, addr);
 }
 
@@ -715,7 +715,7 @@ static int handle_control_get_endpoint_id(struct ctx *ctx, int sd,
 
 	req = (void *)buf;
 	resp->ctrl_hdr.command_code = req->ctrl_hdr.command_code;
-	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP;
+	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP_MASK & req->ctrl_hdr.rq_dgram_inst;
 
 	resp->eid = local_addr(ctx, addr->smctp_ifindex);
 	if (ctx->default_role == ENDPOINT_ROLE_BUS_OWNER)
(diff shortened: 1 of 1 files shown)


src/mctpd.c
```diff
@@ -64,7 +64,7 @@ static mctp_eid_t eid_alloc_max = 0xfe;
 static size_t MAX_PEER_SIZE = 1000000;
 
 static const uint8_t RQDI_REQ = 1 << 7;
-static const uint8_t RQDI_RESP = 0x0;
```
jk-ozlabs, line 67: I'd like to keep the `RQDI_RESP` define here, for symmetry with `RQDI_REQ`. Just add the `_MASK` alongside.

src/mctpd.c
```diff
@@ -832,7 +832,7 @@ static int handle_control_unsupported(struct ctx *ctx, int sd,
 
 	req = (void *)buf;
 	resp->ctrl_hdr.command_code = req->command_code;
-	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP;
+	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP_MASK & req->ctrl_hdr.rq_dgram_inst;
```
jk-ozlabs, line 835: `->ctrl_hdr` is not a member of the `req` control header. Please give the changes at least a quick compile check before submitting.

src/mctpd.c
```diff
@@ -645,7 +645,7 @@ static int handle_control_set_endpoint_id(struct ctx *ctx, int sd,
 	req = (void *)buf;
 
 	resp->ctrl_hdr.command_code = req->ctrl_hdr.command_code;
-	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP;
+	resp->ctrl_hdr.rq_dgram_inst = RQDI_RESP_MASK & req->ctrl_hdr.rq_dgram_inst;
```
jk-ozlabs, line 648: Minor nit: this reads a bit backwards; can you express as `value & mask` ?

Then, you may want to make it explicit that you're handling the Rq bit (even though it is zero), which indicates that it hasn't been accidentally omitted:


```c
resp->ctrl_hdr.rq_dgram_inst = (req->ctrl_hdr.rq_dgram_inst & IID_MASK) | RQDI_RESP;
```

Or, if you prefer being explicit about the IID:


```c
iid = req->ctrl_hdr.rq_dgram_inst & IID_MASK
resp->ctrl_hdr.rq_dgram_inst = iid | RQDI_RESP;
```

jk-ozlabs (changes requested):

khangng-ampere:
While we are at this, thoughts on moving all the header initialization to some common functions? For consistency and less error prone.

```c
// in src/mctp-control-spec.h

static const uint8_t RQDI_REQ = 1 << 7;
static const uint8_t RQDI_RESP = 0x0;
static const uint8_t IID_MASK = 0x1f;

inline void mctp_ctrl_hdr_init_req(struct mctp_ctrl_msg_hdr* req, uint8_t iid, uint8_t command_code)
{
	assert(iid <= IID_MASK);
	req->rq_dgram_inst = iid | RQDI_REQ;
	req->command_code = command_code;
}

inline void mctp_ctrl_hdr_init_resp(struct mctp_ctrl_msg_hdr* resp, struct mctp_ctrl_msg_hdr req)
{
	resp->command_code = req.command_code;
	resp->rq_dgram_inst = (req.rq_dgram_inst & IID_MASK) | RQDI_RESP;
}
```

Usage:

```c
struct mctp_ctrl_resp_get_mctp_ver_support *resp = NULL;
uint8_t respbuf[sizeof(*resp) + 4 * sizeof(*versions)];
resp = (void *)respbuf;
mctp_ctrl_hdr_init_resp(resp, req->ctrl_hdr);
```

For ergonormic, might consider something like

```c
inline void* mctp_ctrl_hdr_init_resp(void* buf, size_t buf_len, struct mctp_ctrl_msg_hdr req)
{
	struct mctp_ctrl_hdr *resp = buf;
	assert(buf_len >= sizeof(*resp));
	resp->command_code = req.command_code;
	resp->rq_dgram_inst = (req.rq_dgram_inst & IID_MASK) | RQDI_RESP;
	return buf;
}
```

Usage:

```c
struct mctp_ctrl_resp_get_mctp_ver_support *resp = NULL;
uint8_t respbuf[sizeof(*resp) + 4 * sizeof(*versions)];
resp = mctp_ctrl_hdr_init_resp(respbuf, sizeof(respbuf), req->ctrl_hdr);
```

jk-ozlabs:
> While we are at this, thoughts on moving all the header initialization to some common functions?

Yep, that would be good, but I would suggest as a separate change.

I'd prefer the first version though (which takes a `struct mctp_ctrl_msg_hdr`), as the caller may already have one allocated (ie, for fixed-size responses).

src/mctpd.c
```diff
@@ -64,7 +64,7 @@ static mctp_eid_t eid_alloc_max = 0xfe;
 static size_t MAX_PEER_SIZE = 1000000;
 
 static const uint8_t RQDI_REQ = 1 << 7;
-static const uint8_t RQDI_RESP = 0x0;
+static const uint8_t RQDI_RESP_MASK = 0x7f;
```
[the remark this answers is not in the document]
gtossk2: Acked.

src/mctpd.c
gtossk2: Acked.

src/mctpd.c
gtossk2: My bad. Fix this and compile successfully.

jk-ozlabs:
OK, the updates look good, but can you squash into a single change? No point in having the initial version present in the tree as a commit.

Let me know if you need a hand with that!

gtossk2:
Sure. Squash into single commit and fix clang-format issue. Thanks!

PR #100, opened 2025-07-31
mctp-bench: Support recv requests

mkj:
"mctp-bench recv" can now optionally request that a peer start sending with a specified message count and message length.

"mctp-bench recv eid 20 len 987 count 10000"

The send equivalent is not yet implemented. Currently mctp-bench will continue listening forever, since it won't necessarily know that all packets have been received (if drops occur). That could be changed in future.

With default arguments "mctp-bench recv" behaviour is unchanged.

### DIFF
--- a/src/mctp-bench.c
+++ b/src/mctp-bench.c
@@ -12,6 +12,10 @@
 #include <sys/socket.h>
 #include <time.h>
 #include <unistd.h>
+#include <linux/prctl.h>
+#include <sys/prctl.h>
+#include <assert.h>
+#include <sys/random.h>
 
 #include "mctp.h"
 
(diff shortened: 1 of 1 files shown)


src/mctp-bench.c
```diff
@@ -350,10 +425,11 @@ static int mctp_bench_send(struct mctp_bench_send_args send_args)
 static void usage(void)
 {
 	fprintf(stderr, "'mctp-bench send' [len <value>] eid [<net>,]<eid>\n");
-	fprintf(stderr, "'mctp-bench recv'\n");
+	fprintf(stderr,
+		"'mctp-bench recv' eid [<net>,]<eid> [len <value>] [count <value>]\n");
```
jk-ozlabs, line 429: The `eid` is optional too, right?

src/mctp-bench.c
```diff
@@ -46,13 +73,43 @@ struct recv_ctx {
...
 static const uint32_t SEQ_START = UINT32_MAX - 5;
 static const uint16_t MAGIC_VAL = 0xbeca;
 static const int DEFAULT_NET = MCTP_NET_ANY;
 static const int DEFAULT_SECONDS_INTERVAL = 2;
 
+static const uint16_t COMMAND_MAGIC = 0x22dd;
+static const uint8_t COMMAND_VERSION = 1;
+
+enum command {
+	COMMAND_RESPONSE = 0x00,
```
jk-ozlabs, line 98: Will responses to all (future) commands use the same opcode?

(I assume the IID will let us catch mismatched responses, but would a request-specific-response type help to diagnnose?)

jk-ozlabs (commented):
Couple of comments inline, but looks good in general.

src/mctp-bench.c
mkj: I was intending to use COMMAND_RESPONSE for all, given IID should be reliable.

src/mctp-bench.c
mkj: I'd fixed that

PR #110, opened 2025-08-29
Minor: fix bridge implementation

faizana-nvidia:
This PR addresses minor fixes for bridge implementation.

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -2840,8 +2840,8 @@ static int peer_route_update(struct peer *peer, uint16_t type)
 			gw_addr.net = peer->net;
 			gw_addr.eid = peer->eid;
 			rc = mctp_nl_route_del(peer->ctx->nl, peer->pool_start,
-					       peer->pool_size - 1,
-					       peer->phys.ifindex, &gw_addr);
+					       peer->pool_size - 1, 0,
+					       &gw_addr);
 			if (rc < 0)
 				warnx("failed to delete route for peer pool eids %d-%d %s",
 				      peer->pool_start,
@@ -4707,12 +4707,35 @@ static int endpoint_allocate_eids(struct peer *peer)
 		warnx("Invalid pool start %d", peer->pool_start);
 		return -1;
 	}
+	/* Add gateway route for all bridge's downstream EIDs.
+	 * After allocation, the endpoint may initiate communication
+	 * immediately, so set up routes for downstream endpoints beforehand.
+	 */
+	struct mctp_fq_addr gw_addr = { 0 };
+	gw_addr.net = peer->net;
+	gw_addr.eid = peer->eid;
+	rc = mctp_nl_route_add(peer->ctx->nl, peer->pool_start,
+			       peer->pool_size - 1, 0, &gw_addr, peer->mtu);
+	if (rc < 0 && rc != -EEXIST) {
+		warnx("Failed to add gateway route for EID %d: %s", gw_addr.eid,
+		      strerror(-rc));
+		return rc;
+	}
+
 	rc = endpoint_send_allocate_endpoint_ids(
 		peer, peer->pool_start, peer->pool_size,
 		mctp_ctrl_cmd_allocate_eids_alloc_eids, &allocated_pool_size,
 		&allocated_pool_start);
 	if (rc) {
 		warnx("Failed to allocate downstream EIDs");
+		// delete prior set routes for downstream endpoints
+		rc = mctp_nl_route_del(peer->ctx->nl, peer->pool_start,
+				       peer->pool_size - 1, 0, &gw_addr);
+		if (rc < 0)
+			warnx("failed to delete route for peer pool eids %d-%d %s",
+			      peer->pool_start,
+			      peer->pool_start + peer->pool_size - 1,
+			      strerror(-rc));
 		//reset peer pool
 		peer->pool_size = 0;
 		peer->pool_start = 0;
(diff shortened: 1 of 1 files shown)


faizana-nvidia:
I'm a bit confused here, while I tested the intended gateway route creation in test framework, that seem to be working as expected if ifindex = 0 as arg in mctp_nl_route_add().

But same fails in my machine.

> Aug 29 07:56:49 hgxb300 mctpd[3669]: Error from kernel: Invalid argument (-22)
> Aug 29 07:56:49 hgxb300 mctpd[3669]:   incorrect format
> Aug 29 07:56:49 hgxb300 mctpd[3669]: mctpd: Failed to add gateway route for EID 10: Invalid argument

though per my understanding of implementation func: ```fill_rtalter_args()``` in mctpd we do need to pass ifindex zero

```
	if (ifindex) {
		msg->nh.nlmsg_len += mctp_put_rtnlmsg_attr(
			&rta, &rta_len, RTA_OIF, &ifindex, sizeof(ifindex));
	} else {
		msg->nh.nlmsg_len += mctp_put_rtnlmsg_attr(
			&rta, &rta_len, RTA_GATEWAY, gw, sizeof(*gw));
	}
```

jk-ozlabs:
The forwarding patches are upstream, but not in a released version as yet. 6.17 should have the merged gateway support.

faizana-nvidia:
> The forwarding patches are upstream, but not in a released version as yet. 6.17 should have the merged gateway support.

Thanks for the infor, yes I can see it in 6.17, will this be a blocker in merge till release happens?

faizana-nvidia:
> No, that shouldn't block us here, as long as you're able to test with a pre-release kernel.

Ack, will update soon on testing.

faizana-nvidia:
I was able to test this.

`mctp route show`
`eid min 11 max 25 net 1 gw 10 mtu 68`
`eid min 10 max 10 net 1 dev mctpusb0 mtu 0`

jk-ozlabs:
On cc7ea36e, you mention:

>  If endpoint responds with some invalid endpoint id range, assumed
> bridge pool space needs to reset too, this is needed as it misleads
> remove_peer() to delete gateway routes too which only be created
> after bridge pool allocation

I think we also have an issue that the pool route is not created until after the allocate endpoint IDs call. This gives us the following race:

| BO | bridge| EP |
|------|----------|-----|
| Sends Allocate Endpoint IDs| | |
| | Allocates downstream endpoint ids ||
| | Sends Set Endpoint ID to EP | |
| | | now has an EID, sends a message to BO |
| **Receives message, cannot reply** | | |
| Adds pool route | | |

which is essentially the same thing fixed in 9711f0bf1b, but with bridges.

faizana-nvidia:
> On [cc7ea36](https://github.com/CodeConstruct/mctp/commit/cc7ea36e4c367ff3e7c6cfeafa288eb0a92984c1), you mention:
> 
> > If endpoint responds with some invalid endpoint id range, assumed
> > bridge pool space needs to reset too, this is needed as it misleads
> > remove_peer() to delete gateway routes too which only be created
> > after bridge pool allocation
> 
> I think we also have an issue that the pool route is not created until after the allocate endpoint IDs call. This gives us the following race:
> 
> BO	bridge	EP
> Sends Allocate Endpoint IDs		
> Allocates downstream endpoint ids	
> Sends Set Endpoint ID to EP	
> now has an EID, sends a message to BO
> **Receives message, cannot reply**		
> Adds pool route		
> which is essentially the same thing fixed in [9711f0b](https://github.com/CodeConstruct/mctp/commit/9711f0bf1b8f47d4d6e3ac58363a5f35ac8c4bd7), but with bridges.

If i understand the origin problem statement correctly which got  fixed in [9711f0b](https://github.com/CodeConstruct/mctp/commit/9711f0bf1b8f47d4d6e3ac58363a5f35ac8c4bd7),,  the issue is right after the moment EID is assigned to an endpoint, we were spending few moments during publishing the peer and then setting up the routes, right at this window, endpoint could very well begin communicating to BusOwner which BusOwner might not be able to respond to since routes doesn't exist yet.

Going by same implication, we should setup routes for downstream eid just before AllocateEndpoint ID and may have delete these routes if Allocation fails.

jk-ozlabs:
> Going by same implication, we should setup routes for downstream eid just before AllocateEndpoint ID and may have delete these routes if Allocation fails.

yes, that's correct.

However, no specific need to have that fixed in this PR; more a note that we'll need to address it at some point.

src/mctpd.c
```diff
@@ -2856,13 +2856,18 @@ static int peer_route_update(struct peer *peer, uint16_t type)
 static int setup_added_peer(struct peer *peer)
 {
 	int rc;
+	struct net *n;
+	n = lookup_net(peer->ctx, peer->net);
```
jk-ozlabs, line 2864: Just do this as a combined declaration and init.

(and preferrably reverse-christmas tree the declarations where possible:

```c
    struct net *n = lookup_net(peer->ctx, peer->net);
    int rc;
```

We'll also need to handle errors on `!n` too.

src/mctpd.c
```diff
@@ -2856,13 +2856,18 @@ static int peer_route_update(struct peer *peer, uint16_t type)
...
 
 	// Set minimum MTU by default for compatibility. Clients can increase
 	// this with .SetMTU as needed
 	peer->mtu = mctp_nl_min_mtu_byindex(peer->ctx->nl, peer->phys.ifindex);
 
-	// add route before querying
-	add_peer_route(peer);
+	// add route before querying for non-bridged endpoints.
+	// bridged endpoints will be routed via gateway route from bridge itself
+	// without need of neighbour
```
jk-ozlabs, line 2872: no need to introduce the neighours into this; just something like

```c
// add route before querying for non-bridged endpoints.
// bridged endpoints will use the bridge's pool range route.
```

src/mctpd.c
```diff
@@ -2083,10 +2083,14 @@ static int endpoint_assign_eid(struct ctx *ctx, sd_bus_error *berr,
 	add_peer_route(peer);
 
 	rc = endpoint_send_set_endpoint_id(peer, &new_eid, &req_pool_size);
-	if (rc == -ECONNREFUSED)
+	if (rc == -ECONNREFUSED) {
 		sd_bus_error_setf(
 			berr, SD_BUS_ERROR_FAILED,
 			"Endpoint returned failure to Set Endpoint ID");
+		peer->pool_size = 0;
```
jk-ozlabs, line 2090: This probably warrants a comment. Perhaps:

```c
// we have not yet created the pool route, reset here so that
// remove_peer() does not attempt to remove it
```

Also: why only on `ECONNREFUSED`, and not in the general `rc < 0` case below?

jk-ozlabs (commented):
Couple of minor things, plus a typo (`birdge`) in the commit message for patch 3.

Otherwise, all looks good.

src/mctpd.c
faizana-nvidia: Ack

src/mctpd.c
faizana-nvidia: Ack

faizana-nvidia:
Addressed the comments, once this merges, will rebase other PR [85](https://github.com/CodeConstruct/mctp/pull/85) to ease the review

PR #123, opened 2025-10-15
mctpd: replace sd-event timers

khangng-ampere:
In tests, replace all sd_event timer usages with trio.testing.MockClock
backed I/O sources.

Because we still have real timeouts (D-Bus call to mctpd subprocess,
mctpd SO_RCVTIMEO wait, ...), autojump_threshold (how much to wait
before skipping Trio timeouts) is set to a reasonable value to take that
into account.

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -497,8 +497,9 @@ static int wait_fd_timeout(int fd, short events, uint64_t timeout_usec)
 	if (rc < 0)
 		goto out;
 
-	rc = sd_event_add_time_relative(ev, NULL, CLOCK_MONOTONIC, timeout_usec,
-					0, cb_exit_loop_timeout, NULL);
+	rc = mctp_ops.sd_event.add_time_relative(ev, NULL, CLOCK_MONOTONIC,
+						 timeout_usec, 0,
+						 cb_exit_loop_timeout, NULL);
 	if (rc < 0)
 		goto out;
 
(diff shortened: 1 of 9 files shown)


khangng-ampere:
This PR is still missing some crucial pieces:

- Communicate to the callback timer what is the next time to wait for.
- Calling the timer entrypoint from the test runner. If we can, we should be able to use all the [trio timers and time travel functions](https://trio.readthedocs.io/en/stable/reference-testing.html) for free.

src/mctpd.c
```diff
@@ -3115,23 +3073,13 @@ static int method_endpoint_remove(sd_bus_message *call, void *data,
...
 
-	/*
-	 * Error handling policy:
-	 *
-	 * 1. Any resource management error prior to Treclaim is handled by
-	 *    rescheduling the poll query, unless it is scheduling the poll
-	 *    query itself that fails.
-	 *
-	 * 2. If scheduling the poll query fails then the endpoint is removed.
-	 */
```
khangng-ampere: I believe this comment is for when `sd_event_source_*` fails. It does not applies anymore, so I removed it.

jk-ozlabs:
I can't say I'm *super* keen on using timerfds. Constraining complexity in the production code takes priority over test implementations, for me.

I'll do some experimentation and see if we can mock with the standard sd_event time handling.

jk-ozlabs:
I think this looks better, but would like to see if we can decouple the test suite timer from mctpd's own concept of time. This would mean replacing the trio autojump timer with one that we could explicitly advance in the test cases.

The challenge there is syncronisation between the advance and the registration of a new timer though...

jk-ozlabs:
What's the general plan here? The final HAVE_LIBSYSTEMD change looks like a temporary fix - or is that something you're planning to include as-is?

jk-ozlabs:
Any thoughts on something like this: https://github.com/jk-ozlabs/mctp/commit/e6916477f1dd698d87d585efb43fdc5b14d29594 ?

jk-ozlabs:
Yep, all good there. Thank you for the contribution!

PR #125, opened 2025-10-27
Fix mctp-bench build on musl libc

jk-ozlabs:
Remove the `prctl.h` includes, as they're unused. Do a bit of a header tidy at the same time.


kraj (commented):
LGTM

PR #128, opened 2025-10-29
mctpd: implement routing table polling

khangng-ampere:
As [discussed on OpenBMC Discord](https://discord.com/channels/775381525260664832/778790638563885086/1282947932093415446), according to the DSP0236 spec,
there is no good way to discover all active endpoints aside from
polling/scanning the entire network.

This adds an initial routing table polling mechanism. For each bridge,
mctpd will send Get Routing Table and add the endpoints to D-Bus.
If some of the endpoints are bridges, mctpd recursively enables
polling for those bridges.

To kick start the process, mctpd will enable polling on its bus owner,
discovered when mctpd receives the Set Endpoint ID message.

mctp.routing_table_polling_interval_ms config can be used to tweak the
delay between pollings.

---

I have just rebased this patch on top of my other patches. Thought I should also open a PR for it too.

It is still a bit rough and depends on other PRs so I am marking this as draft.

### DIFF
--- a/conf/mctpd.conf
+++ b/conf/mctpd.conf
@@ -5,6 +5,9 @@ mode = "bus-owner"
 [mctp]
 message_timeout_ms = 30
 
+# Delay between each Get Routing Table message.
+routing_table_polling_interval_ms = 1000
+
 # Specify a UUID: not generally required - mctpd will query the system UUID
 # where available.
 # uuid = "21f0f554-7f7c-4211-9ca1-6d0f000ea9e7"
--- a/src/mctp-control-spec.h
+++ b/src/mctp-control-spec.h
@@ -413,3 +413,11 @@ enum mctp_phys_binding {
 	MCTP_PHYS_BINDING_UCIE = 0x09,
 	MCTP_PHYS_BINDING_VENDOR = 0xFF,
 };
+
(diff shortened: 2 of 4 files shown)


khangng-ampere:
> There was some earlier discussion that we don't have reliable enough information in Get Routing Table Entries to assume that downstream endpoints are present.
>
> So, we'll probably need to discuss the approach first. The linked Discord chat was some time ago :)
> 
> As a summary of my thoughts here: Get Routing Table Entries would be more efficient than individual-endpoint polling, but I'm not convinced we get accurate enough data from that. For example, a bridge may indicate that is has a downstream route to a range of EIDs (through data in the GRTE), but those individual endpoints may not exist at all.

As for what I am thinking, it is indeed not accurate, but such is also the nature of the endpoint objects published on D-Bus. In normal operation, even without this PR (or the other one), some of the endpoints can go offline, and the D-Bus endpoint objects can get outdated/inaccurate. Applications/Consumers of the D-Bus list also already had to deal with that using the `.Recovery` mechanism, so I think it should apply here as well.

> Also, given mctpd does not implement GRTE commands itself, I figure we'll need some infrastructure for that too.

Agreed.

PR #131, opened 2025-11-28
mctp-netlink: Cleanups from a -Db_sanitize=address,undefined build run in a target environment

amboar:
I was doing some testing a while back and found a couple of minor issues. Rediscovered them today.


jk-ozlabs:
Nice. Should we be doing some UBSANning in CI?

jk-ozlabs:
aww, we can't pass NULL even when we're representing the size as zero?

PR #141, opened 2026-03-03
mctpd: update supports_vdm check

Ruta01:
supports_vdm check to be decided based on supported PCIE/IANA message type instead of formatID.
Fixes: cb182be ("mctpd: Add VendorDefinedMessageTypes property")

### DIFF
--- a/tests/test_mctpd.py
+++ b/tests/test_mctpd.py
@@ -665,6 +665,37 @@ async def test_query_vdm_types(dbus, mctpd):
     assert set(vdm_types) == set(ret_vdm_types)
 
 
+async def test_query_vdm_types_no_control(dbus, mctpd):
+    """Test that we query VDM types if *only* a VDM type is reported in the
+    non-vendor Message Type Support response
+    """
+    iface = mctpd.system.interfaces[0]
+    vdm_types = [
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
Also, you may want to check you git author/commiter settings; it looks like you have a username there instead of a full name.

Perhaps:

```
git config --global user.name "Ruta Abhyankar"
```

\- but if the `rutaabhyankar` format is inetentional, that's OK too.

jk-ozlabs:
Thanks for the update. The failing test is due to the actual bug; I have some testcase changes that I will include in the merge.

Ruta01:
> Thanks for the update. The failing test is due to the actual bug; I have some testcase changes that I will include in the merge.

I updated the test. Looks like some format issue in test file. Will check.

jk-ozlabs:
Don't worry about the test changes: I have some reworks done on how the VDM types are handled within the Endpoint code, which I will merge alongside your change.

Might be helpful to add a Fixes line to your commit though:

```
Fixes: cb182be ("mctpd: Add VendorDefinedMessageTypes property")
```

jk-ozlabs:
OK, I have pushed your branch to my full set of changes (including your fix, and a test for this case). Please let me know if that looks OK to you.

PR #145, opened 2026-03-16
mctpd: initialize peer to NULL to fix error path

haiyuewa:
Initialize peer to NULL to avoid using an uninitialized pointer in the error path when sd_bus_message_read() fails to get 'eid'.


jk-ozlabs:
Merged, thank you!

PR #148, opened 2026-03-19
mctpd: add AssignBridgeStatic method

potinlai:
Add AssignBridgeStatic method to assign static EID with bridge pool allocation.

Tested:
```
root@bmc:~# busctl call au.com.codeconstruct.MCTP1 /au/com/codeconstruct/mctp1/interfaces/mcu1u2u1u4u1 au.com.codeconstruct.MCTP.BusOwner1 AssignBridgeStatic ayyyy 0 30 5 31
yisb 30 1 "/au/com/codeconstruct/mctp1/networks/1/endpoints/30" true
root@bmc:~# busctl introspect au.com.codeconstruct.MCTP1 /au/com/codeconstruct/mctp1/networks/1/endpoints/30
NAME                                TYPE      SIGNATURE RESULT/VALUE                           FLAGS
au.com.codeconstruct.MCTP.Bridge1   interface -         -                                      -
.PoolEnd                            property  y         35                                     const
.PoolStart                          property  y         31                                     const
au.com.codeconstruct.MCTP.Endpoint1 interface -         -                                      -
.Recover                            method    -         -                                      -
.Remove                             method    -         -                                      -
.SetMTU                             method    u         -                                      -
.Connectivity                       property  s         "Available"                            emits-change
org.freedesktop.DBus.Introspectable interface -         -                                      -
.Introspect                         method    -         s                                      -
org.freedesktop.DBus.Peer           interface -         -                                      -
.GetMachineId                       method    -         s                                      -
.Ping                               method    -         -                                      -
org.freedesktop.DBus.Properties     interface -         -                                      -
.Get                                method    ss        v                                      -
.GetAll                             method    s         a{sv}                                  -
.Set                                method    ssv       -                                      -
.PropertiesChanged                  signal    sa{sv}as  -                                      -
xyz.openbmc_project.Common.UUID     interface -         -                                      -
.UUID                               property  s         "8b7e3d86-5fa5-298c-a9d6-a2d5d4baa6fa" const
xyz.openbmc_project.MCTP.Endpoint   interface -         -                                      -
.EID                                property  y         30                                     const
.NetworkId                          property  u         1                                      const
.SupportedMessageTypes              property  ay        4 1 5 126 127                          const
.VendorDefinedMessageTypes          property  a(yvq)    0                                      const
```

Signed-off-by: Potin Lai <potin.lai@quantatw.com>

### DIFF
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -203,6 +203,37 @@ This call will fail if the endpoint already has an EID, and that EID is
 different from `static-EID`, or if `static-EID` is already assigned to another
 endpoint.
 
+#### `.AssignBridgeStatic`: `ayyyy` → `yyisb`
+
+This method performs a static assignment with bridge EID pool allocation.
+
+Similar to AssignEndpointStatic, but takes an additional pool-start-EID
+and pool-size argument:
+
+```
+AssignBridgeStatic <hwaddr> <static-EID> <pool-start-EID> <pool-size>
+```
+
+`<pool-start-EID>`: The starting EID for the range of EIDs being allocated.
+This EID must be contiguous with the bridge’s own EID, to comply with the
+current MCTP 1.3.x specification. In future releases, this restriction
+may be relaxed.
+If passed as 0, mctpd will automatically choose a pool start EID.
+
+`<pool-size>`: The number of EIDs to be allocated to the bridge's EID pool.
+
+Returns:
+```
+eid        (byte)
+poolstart  (byte)
+net        (integer)
+path       (string)
+new        (bool) - true if a new EID was assigned
+```
+
+Where `eid` is the assigned EID, `poolstart` is the starting EID of the allocated
+bridge EID pool, and the remaining fields are the same as the methods above.
+
 #### `.LearnEndpoint`: `ay` → `yisb`
 
 Like SetupEndpoint but will not assign EIDs, will only query endpoints for a
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -2329,6 +2329,34 @@ static int allocate_eid(struct ctx *ctx, struct net *net,
 	return -1;
 }
 
+/**
+ * Validate EID pool allocation for an MCTP bridge.
+ *
+ * Ensures the pool starts at bridge_eid + 1 and verifies that neither the
+ * bridge EID nor the pool range conflicts with existing peer allocations in
+ * the network.
+ *
+ * @return 0 on success, -1 on failure.
+ */
+static int check_bridge_eid_alloc(struct net *net, mctp_eid_t bridge_eid,
+				  mctp_eid_t pool_start, unsigned int pool_size)
+{
+	mctp_eid_t eid;
+
+	if (pool_start != (bridge_eid + 1))
+		return -1;
+
+	if (net->peers[bridge_eid])
+		return -1;
+
+	for (eid = pool_start; eid < (pool_start + pool_size); eid++) {
+		if (net->peers[eid])
+			return -1;
+	}
+
+	return 0;
+}
+
 static int endpoint_assign_eid(struct ctx *ctx, sd_bus_error *berr,
 			       const dest_phys *dest, struct peer **ret_peer,
 			       mctp_eid_t static_eid, bool assign_bridge)
@@ -2427,6 +2455,100 @@ static int endpoint_assign_eid(struct ctx *ctx, sd_bus_error *berr,
 				berr, SD_BUS_ERROR_FAILED,
 				"Endpoint requested EID %d instead of assigned %d, already used",
 				new_eid, peer->eid);
+		}
+		if (rc < 0) {
+			remove_peer(peer);
+			return rc;
+		}
+	}
+
+	rc = setup_added_peer(peer);
+	if (rc < 0)
+		return rc;
+	*ret_peer = peer;
+
+	return 0;
+}
+
+static int
+endpoint_assign_bridge_static_eid(struct ctx *ctx, sd_bus_error *berr,
+				  const dest_phys *dest, struct peer **ret_peer,
+				  mctp_eid_t static_eid, mctp_eid_t pool_start,
+				  uint8_t pool_size)
+{
+	struct peer *peer = NULL;
+	uint8_t req_pool_size;
+	struct net *n = NULL;
+	uint32_t net;
+	int rc;
+
+	net = mctp_nl_net_byindex(ctx->nl, dest->ifindex);
+	if (!net) {
+		bug_warn("No net known for ifindex %d", dest->ifindex);
+		return -EPROTO;
+	}
+
+	n = lookup_net(ctx, net);
+	if (!n) {
+		bug_warn("Unknown net %d", net);
+		return -EPROTO;
+	}
+
+	rc = check_bridge_eid_alloc(n, static_eid, pool_start, pool_size);
+	if (rc) {
+		warnx("Cannot allocate static EID (+pool %d) on net %d for %s",
+		      pool_size, net, dest_phys_tostr(dest));
+		sd_bus_error_setf(berr, SD_BUS_ERROR_FAILED, "Ran out of EIDs");
+		return -EADDRNOTAVAIL;
+	}
+
+	rc = add_peer(ctx, dest, static_eid, net, &peer, false);
+	if (rc < 0)
+		return rc;
+
+	peer->pool_size = pool_size;
+	if (peer->pool_size)
+		peer->pool_start = pool_start;
+
+	add_peer_route(peer);
+
+	rc = endpoint_send_set_endpoint_id(peer, &static_eid, &req_pool_size);
+	if (rc == -ECONNREFUSED)
+		sd_bus_error_setf(
+			berr, SD_BUS_ERROR_FAILED,
+			"Endpoint returned failure to Set Endpoint ID");
+
+	if (rc < 0) {
+		// we have not yet created the pool route, reset here so that
+		// remove_peer() does not attempt to remove it
+		peer->pool_size = 0;
+		peer->pool_start = 0;
+		remove_peer(peer);
+		return rc;
+	}
+
+	if (req_pool_size > peer->pool_size) {
+		warnx("EID %d: requested pool size (%d) > pool size available (%d), limiting.",
+		      peer->eid, req_pool_size, peer->pool_size);
+	} else {
+		// peer will likely have requested less than the available range
+		peer->pool_size = req_pool_size;
+	}
+
+	if (!peer->pool_size)
+		peer->pool_start = 0;
+
+	if (static_eid != peer->eid) {
+		// avoid allocation for any different EID in response
+		warnx("Mismatch of requested from received EID, resetting the pool");
+		peer->pool_size = 0;
+		peer->pool_start = 0;
+		rc = change_peer_eid(peer, static_eid);
+		if (rc == -EEXIST) {
+			sd_bus_error_setf(
+				berr, SD_BUS_ERROR_FAILED,
+				"Endpoint requested EID %d instead of assigned %d, already used",
+				static_eid, peer->eid);
 		}
 		if (rc < 0) {
 			remove_peer(peer);
@@ -3121,6 +3243,98 @@ static int method_assign_endpoint_static(sd_bus_message *call, void *data,
 
 	return sd_bus_reply_method_return(call, "yisb", peer->eid, peer->net,
 					  peer_path, 1);
+err:
+	set_berr(ctx, rc, berr);
+	return rc;
+}
+
+static int method_assign_bridge_static(sd_bus_message *call, void *data,
+				       sd_bus_error *berr)
+{
+	dest_phys desti, *dest = &desti;
+	const char *peer_path = NULL;
+	struct peer *peer = NULL;
+	struct link *link = data;
+	struct ctx *ctx = link->ctx;
+	uint8_t eid;
+	uint8_t pool_start;
+	uint8_t pool_size;
+	bool new = true;
+	int rc;
+
+	dest->ifindex = link->ifindex;
+	if (dest->ifindex <= 0)
+		return sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+					 "Unknown MCTP interface");
+
+	rc = message_read_hwaddr(call, dest);
+	if (rc < 0)
+		goto err;
+
+	rc = sd_bus_message_read(call, "y", &eid);
+	if (rc < 0)
+		goto err;
+
+	rc = sd_bus_message_read(call, "y", &pool_start);
+	if (rc < 0)
+		goto err;
+
+	rc = sd_bus_message_read(call, "y", &pool_size);
+	if (rc < 0)
+		goto err;
+
+	rc = validate_dest_phys(ctx, dest);
+	if (rc < 0)
+		return sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+					 "Bad physaddr");
+
+	if (pool_start == 0) {
+		pool_start = eid + 1;
+	} else if (pool_start != (eid + 1)) {
+		return sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+					 "Bad pool start EID");
+	}
+
+	peer = find_peer_by_phys(ctx, dest);
+	if (peer) {
+		if (eid && peer->eid == eid && pool_size == peer->pool_size) {
+			new = false;
+			goto out;
+		}
+
+		remove_peer(peer);
+		peer = NULL;
+	} else {
+		uint32_t netid;
+
+		// is the requested EID already in use? if so, reject
+		netid = mctp_nl_net_byindex(ctx->nl, dest->ifindex);
+		peer = find_peer_by_addr(ctx, eid, netid);
+		if (peer) {
+			return sd_bus_error_setf(berr,
+						 SD_BUS_ERROR_INVALID_ARGS,
+						 "Address in use");
+		}
+	}
+
+	rc = endpoint_assign_bridge_static_eid(ctx, berr, dest, &peer, eid,
+					       pool_start, pool_size);
+	if (rc < 0) {
+		goto err;
+	}
+
+	if (peer->pool_size > 0)
+		endpoint_allocate_eids(peer);
+
+out:
+	peer_path = path_from_peer(peer);
+	if (!peer_path)
+		goto err;
+
+	return sd_bus_reply_method_return(call, "yyisb", peer->eid,
+					  peer->pool_start, peer->net,
+					  peer_path, new);
+
 err:
 	set_berr(ctx, rc, berr);
 	return rc;
@@ -4053,6 +4267,21 @@ static const sd_bus_vtable bus_link_owner_vtable[] = {
 		SD_BUS_PARAM(path)
 		SD_BUS_PARAM(new),
 		method_assign_endpoint_static,
+		0),
+
+	SD_BUS_METHOD_WITH_NAMES("AssignBridgeStatic",
+		"ayyyy",
+		SD_BUS_PARAM(physaddr)
+		SD_BUS_PARAM(eid)
+		SD_BUS_PARAM(poolstart)
+		SD_BUS_PARAM(poolsize),
+		"yyisb",
+		SD_BUS_PARAM(eid)
+		SD_BUS_PARAM(poolstart)
+		SD_BUS_PARAM(net)
+		SD_BUS_PARAM(path)
+		SD_BUS_PARAM(new),
+		method_assign_bridge_static,
 		0),
 
 	SD_BUS_METHOD_WITH_NAMES("LearnEndpoint",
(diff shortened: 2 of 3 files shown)


jk-ozlabs:
I'm going to need more context than that commit message. What is the aim here?

src/mctpd.c
```diff
@@ -2288,38 +2288,33 @@ static int endpoint_assign_eid(struct ctx *ctx, sd_bus_error *berr,
...
+		      alloc_size, net, dest_phys_tostr(dest));
+		sd_bus_error_setf(berr, SD_BUS_ERROR_FAILED, "Ran out of EIDs");
+		return -EADDRNOTAVAIL;
+	}
 
+	if (static_eid) {
+		new_eid = static_eid;
+	} else {
 		new_eid = alloc.start;
+	}
```
jk-ozlabs: You can't just redefine the allocated range like this, those EIDs at [static_eid, static_eid + alloc.extent] may not be available.

potinlai: Hi @jk-ozlabs,
Thank you for your suggestion in Discord.
I have updated the change to introduce a new method called `AssignBridgeStatic`.

src/mctpd.c
```diff
@@ -2378,6 +2378,92 @@ static int endpoint_assign_eid(struct ctx *ctx, sd_bus_error *berr,
...
+		return -EPROTO;
+	}
+
+	rc = add_peer(ctx, dest, static_eid, net, &peer, false);
+	if (rc < 0)
+		return rc;
+
+	peer->pool_size = pool_size;
+	if (peer->pool_size)
+		peer->pool_start = pool_start;
```
jk-ozlabs, line 2511: How are you ensuring that this does not conflict with existing allocations?

Also, check that pool_start == static_eid + 1 here, in order to be compliant with the spec.

docs/mctpd.md
```diff
@@ -203,6 +203,21 @@ This call will fail if the endpoint already has an EID, and that EID is
 different from `static-EID`, or if `static-EID` is already assigned to another
 endpoint.
 
+#### `.AssignBridgeStatic`: `ayy` → `yisb`
```
jk-ozlabs: This argument signature does not match the implementation

docs/mctpd.md
```diff
@@ -203,6 +203,21 @@ This call will fail if the endpoint already has an EID, and that EID is
...
+
+Similar to AssignEndpointStatic, but takes an additional pool-size and
+pool-start-EID argument:
+
+```
+AssignBridgeStatic <hwaddr> <static-EID> <pool-size> <pool-start-EID>
```
jk-ozlabs: I think this may read a little better with the two EIDs in order:

```
AssignBridgeStatic <hwaddr> <static-EID> <pool-start-EID> <pool-size>
```

docs/mctpd.md
```diff
@@ -203,6 +203,21 @@ This call will fail if the endpoint already has an EID, and that EID is
...
+```
+
+`<pool-size>`: The number of EIDs to be allocated to the bridge's EID pool.
+
+`<pool-start-EID>`: The starting EID for the range of EIDs being allocated.
```
jk-ozlabs, line 217: We will need to document the requirement that this EID is contiguous with the bridge EID.

src/mctpd.c
```diff
@@ -3071,6 +3157,93 @@ static int method_assign_endpoint_static(sd_bus_message *call, void *data,
...
+	peer = find_peer_by_phys(ctx, dest);
+	if (peer) {
+		if (peer->eid != eid) {
+			return sd_bus_error_setf(
+				berr, SD_BUS_ERROR_INVALID_ARGS,
+				"Already assigned a different EID");
+		}
+
+		// Return existing record.
+		peer_path = path_from_peer(peer);
```
jk-ozlabs: Is this valid here, and is the check sufficient? The existing peer may not be the bridge you expect...

jk-ozlabs (changes requested):
A few things inline, and the commit message needs more explanation. Rather than a chunk of 'Tested' output, can you include some details about the use case and what is implemented here?

jk-ozlabs:
Also, could you add some test cases here? We want to check the successful call, but also conflicts with existing EID allocations.

docs/mctpd.md
potinlai: update.

docs/mctpd.md
potinlai: updated.

docs/mctpd.md
potinlai: updated.

potinlai:
> Also, could you add some test cases here? We want to check the successful call, but also conflicts with existing EID allocations.

Hi @jk-ozlabs,
I’m still trying to understand how to run the test cases. I copied the test-mctpd binary to the BMC and executed it, but I encountered the following error:
```
root@bmc:~# /tmp/test-mctpd
test-mctpd: No MCTP_TEST_SOCK fd provided
```  

Could you please advise on the correct way to run this test?

jk-ozlabs:
The binary doesn't need to run on the BMC; the idea behind the test infrastructure is that the `test-mctpd` build does not have any dependencies on the system MCTP stack (and is designed to run where there is no MCTP capabilities on the system)

The [Testing section](https://github.com/CodeConstruct/mctp#testing) in the README has some details, but you just run the tests locally using something like:

```shell
$ python3 -m venv env
$ ./env/bin/pip install -r tests/requirements.txt
$ export PATH=$PWD/env/bin:$PATH
$ meson setup obj
$ ninja -C obj test
```

docs/mctpd.md
```diff
@@ -203,6 +203,22 @@ This call will fail if the endpoint already has an EID, and that EID is
...
+
+Similar to AssignEndpointStatic, but takes an additional pool-size and
+pool-start-EID argument:
+
+```
+AssignBridgeStatic <hwaddr> <static-EID> <pool-start-EID> <pool-size>
+```
+
+`<pool-start-EID>`: The starting EID for the range of EIDs being allocated.
+This EID must be contiguous with the bridge’s own EID.
```
faizana-nvidia: Hello @jk-ozlabs,

Has there been any plan to consider allocation of pool with non contiguous pool start (w.r.t bridge's own EID)? may be while moving to v.14 DSP0236?
https://github.com/CodeConstruct/mctp/pull/71#issuecomment-3006826382
https://github.com/DMTF/PMCI-WG/issues/1540

jk-ozlabs: It's something I would like to support, but no specific plans at the moment.

Note that this would only be possible after confirming the MCTP implementation version of the bridge device though - which may complicate the external API, since the success/failure of a non-contiguous allocation call would depend on the remote endpoint version. So, it's probably not a big win at the moment, as we already implement the contiguity restriction anyway.

faizana-nvidia: Thank you for the clarification, I suppose this PR will give option to choose variable pool size allocation incase ask is to have pool less/equal to what bridge pool size is (response from SET ENDPOINT ID)

jk-ozlabs: This does bring up a good point though - the new API (introduced by this change) would be somewhat difficult to use when we want to implement split pools, as the caller may not know in advance if the call is valid.

I see a couple of options:

1. We remove the `pool-start` argument, and let mctpd specify the pool start address; or
2. We allow a 0 value for `pool-start`, indicating that mctpd can put the pool range whereever it likes

The downside of (1) is that callers may require the pool range to be predefined, for whatever reason.

@potinlai can you provide some details on your need for the static EID allocation, perhaps? Do you need the pool range to be static, or only the bridge EID?

potinlai: Sorry for the late response.

In my use case, I used to provide a selected range for bridge and pool assignment.
For example, if I want to assign `30` to the bridge and `31–39` to the bridge pool, the call would be:
`AssignBridgeStatic 0 30 31 9`
However, in most cases, the bridge requires fewer resources than the size I assigned.

potinlai:
> Also, could you add some test cases here? We want to check the successful call, but also conflicts with existing EID allocations.

test cases updated.

jk-ozlabs:
Just traveling at the moment, but I'll take a look at the updates shortly. Thanks!

jk-ozlabs:
There was a pending question [above](https://github.com/CodeConstruct/mctp/pull/148#discussion_r3151168724):

> @potinlai can you provide some details on your need for the static EID allocation, perhaps? Do you need the pool range to be static, or only the bridge EID?

Could you give us some details for your use-case? particularly around the requirement for pool addressing.

This will dictate what future changes might be applicable, and how we should structure the interface to suit those.

potinlai:
> There was a pending question [above](https://github.com/CodeConstruct/mctp/pull/148#discussion_r3151168724):
> 
> > @potinlai can you provide some details on your need for the static EID allocation, perhaps? Do you need the pool range to be static, or only the bridge EID?
> 
> Could you give us some details for your use-case? particularly around the requirement for pool addressing.
> 
> This will dictate what future changes might be applicable, and how we should structure the interface to suit those.

There are 7 SMA devices in system, and each SMA has its own downstream devices.
The SMA's static EID and EID pool allocation as below.
- EID 10 : 11
- EID 20 : 21
- EID 30 : 31-35
- EID 40 : 41-45
- EID 50 : 51-55
- EID 60 : 61-65
- EID 70 : 71-72

Below is MCTP tree on my BMC.
```
root@bmc:~# busctl tree au.com.codeconstruct.MCTP1 
└─ /au
  └─ /au/com
    └─ /au/com/codeconstruct
      └─ /au/com/codeconstruct/mctp1
        ├─ /au/com/codeconstruct/mctp1/interfaces
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/lo
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mctpi2c13
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mcu1u1u4u1
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mcu1u2u1u1
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mcu1u2u1u3u1
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mcu1u2u1u4u1
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mcu1u2u2u1
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mcu1u2u2u3u1
        │ ├─ /au/com/codeconstruct/mctp1/interfaces/mcu1u2u2u4u1
        │ └─ /au/com/codeconstruct/mctp1/interfaces/mcu1u3u2
        └─ /au/com/codeconstruct/mctp1/networks
          └─ /au/com/codeconstruct/mctp1/networks/1
            └─ /au/com/codeconstruct/mctp1/networks/1/endpoints
              ├─ /au/com/codeconstruct/mctp1/networks/1/endpoints/10
              ├─ /au/com/codeconstruct/mctp1/networks/1/endpoints/20
              ├─ /au/com/codeconstruct/mctp1/networks/1/endpoints/30
              ├─ /au/com/codeconstruct/mctp1/networks/1/endpoints/40
              ├─ /au/com/codeconstruct/mctp1/networks/1/endpoints/50
              ├─ /au/com/codeconstruct/mctp1/networks/1/endpoints/60
              ├─ /au/com/codeconstruct/mctp1/networks/1/endpoints/70
              └─ /au/com/codeconstruct/mctp1/networks/1/endpoints/8
```


And below is introspect details of EID 30 .
```
root@bmc:~# busctl introspect au.com.codeconstruct.MCTP1 /au/com/codeconstruct/mctp1/networks/1/endpoints/30                                  
NAME                                TYPE      SIGNATURE RESULT/VALUE                           FLAGS       
au.com.codeconstruct.MCTP.Bridge1   interface -         -                                      -           
.PoolEnd                            property  y         35                                     const
.PoolStart                          property  y         31                                     const
au.com.codeconstruct.MCTP.Endpoint1 interface -         -                                      -           
.Recover                            method    -         -                                      -           
.Remove                             method    -         -                                      -           
.SetMTU                             method    u         -                                      -           
.Connectivity                       property  s         "Available"                            emits-change
org.freedesktop.DBus.Introspectable interface -         -                                      -           
.Introspect                         method    -         s                                      -           
org.freedesktop.DBus.Peer           interface -         -                                      -           
.GetMachineId                       method    -         s                                      -           
.Ping                               method    -         -                                      -           
org.freedesktop.DBus.Properties     interface -         -                                      -           
.Get                                method    ss        v                                      -           
.GetAll                             method    s         a{sv}                                  -           
.Set                                method    ssv       -                                      -           
.PropertiesChanged                  signal    sa{sv}as  -                                      -           
xyz.openbmc_project.Common.UUID     interface -         -                                      -           
.UUID                               property  s         "8b7e3d86-5fa5-298c-a9d6-a2d5d4baa6fa" const
xyz.openbmc_project.MCTP.Endpoint   interface -         -                                      -           
.EID                                property  y         30                                     const
.NetworkId                          property  u         1                                      const
.SupportedMessageTypes              property  ay        4 1 5 126 127                          const
```

jk-ozlabs:
Thanks, but I was more after your current requirements around the EID allocation pool.

Particularly: do you require the pool to be static?

It will need to be at the moment, but future mctp spec versions will relax the requirement that the pool is contiguous with the bridge EID. In that case, the caller may choose a non-contiguous pool address, or mctpd may be able to choose a pool range that is free. The former is a bit tricky, as the caller has no idea whether the remote endpoint would support non-contiguous pools, so I suspect the latter option would be more popular.

To accommodate that, I think we would want to allow the `pool_start` argument to be zero, to indicate that mctpd can choose a pool start. And then the caller would need some way to return the chosen pool start address - so we would need an extra `y` output argument for that.

potinlai:
> Thanks, but I was more after your current requirements around the EID allocation pool.
> 
> Particularly: do you require the pool to be static?

Yes, In my use case, I expected the pool is static arranged, so each downstream devices will have same EID assigned.

> 
> It will need to be at the moment, but future mctp spec versions will relax the requirement that the pool is contiguous with the bridge EID. In that case, the caller may choose a non-contiguous pool address, or mctpd may be able to choose a pool range that is free. The former is a bit tricky, as the caller has no idea whether the remote endpoint would support non-contiguous pools, so I suspect the latter option would be more popular.
> 
> To accommodate that, I think we would want to allow the `pool_start` argument to be zero, to indicate that mctpd can choose a pool start. 

That’s a great point about allowing pool_start to be zero for mctpd-managed allocation.
However, I haven't verified if the devices (SMA) on our system support non-contiguous pools.
Could this feature be a seperate enhancement on top of the current changes?

> And then the caller would need some way to return the chosen pool start address - so we would need an extra `y` output argument for that.

I can add extra `y` output.
Should it be `yyisb`? The 2nd `y` for final pool_start eid.

jk-ozlabs:
> However, I haven't verified if the devices (SMA) on our system support non-contiguous pools.
> Could this feature be a seperate enhancement on top of the current changes?

It is unlikely that they support non-contiguous pools, the spec change is quite recent. I am not proposing that we implement that at this stage - only that the APIs we introduce now would be able to support it.

So, the only behaviour we need now is for the pool-start argument, if passed as zero, is set to the bridge-eid+1. Everything else can be a later change.

> I can add extra y output.
> Should it be yyisb? The 2nd y for final pool_start eid.

Sounds great.

potinlai:
> > However, I haven't verified if the devices (SMA) on our system support non-contiguous pools.
> > Could this feature be a seperate enhancement on top of the current changes?
> 
> It is unlikely that they support non-contiguous pools, the spec change is quite recent. I am not proposing that we implement that at this stage - only that the APIs we introduce now would be able to support it.
> 
> So, the only behaviour we need now is for the pool-start argument, if passed as zero, is set to the bridge-eid+1. Everything else can be a later change.
> 
> > I can add extra y output.
> > Should it be yyisb? The 2nd y for final pool_start eid.
> 
> Sounds great.

Thank you for the feedback.
I have updated the change as requested.

docs/mctpd.md
```diff
@@ -203,6 +203,23 @@ This call will fail if the endpoint already has an EID, and that EID is
 different from `static-EID`, or if `static-EID` is already assigned to another
 endpoint.
 
+#### `.AssignBridgeStatic`: `ayyyy` → `yyisb`
+
+Similar to AssignEndpointStatic, but takes an additional pool-size and
+pool-start-EID argument:
```
jk-ozlabs: I think we need a higher-level description of this first. You have included a suitable description below ("This method performs a static assignment with bridge EID pool allocation."), I would just suggest putting that as the first line here.

docs/mctpd.md
```diff
@@ -203,6 +203,23 @@ This call will fail if the endpoint already has an EID, and that EID is
...
+```
+
+`<pool-start-EID>`: The starting EID for the range of EIDs being allocated.
+This EID must be contiguous with the bridge’s own EID.
+If passed as 0, the daemon will automatically set this to `<static-EID>` + 1.
+
+`<pool-size>`: The number of EIDs to be allocated to the bridge's EID pool.
+
+This method performs a static assignment with bridge EID pool allocation.
+
```
jk-ozlabs, line 236: We will also need a description of the `yyisb` outputs. Perhaps just document the assigned EID and pool start outputs, and that `isb` is the same as the functions above.

src/mctpd.c
```diff
@@ -2442,6 +2405,100 @@ static int endpoint_assign_eid(struct ctx *ctx, sd_bus_error *berr,
...
+endpoint_assign_bridge_static_eid(struct ctx *ctx, sd_bus_error *berr,
+				  const dest_phys *dest, struct peer **ret_peer,
+				  mctp_eid_t static_eid, mctp_eid_t pool_start,
+				  uint8_t pool_size)
+{
+	struct net *n = NULL;
+	struct peer *peer = NULL;
+	uint8_t req_pool_size;
+	uint32_t net;
+	int rc;
```
jk-ozlabs: Super minor, but if you're reworking, can you reverse-christmas-tree these?

src/mctpd.c
```diff
@@ -2442,6 +2405,100 @@ static int endpoint_assign_eid(struct ctx *ctx, sd_bus_error *berr,
...
+		remove_peer(peer);
+		return rc;
+	}
+
+	if (req_pool_size > peer->pool_size) {
+		warnx("EID %d: requested pool size (%d) > pool size available (%d), limiting.",
+		      peer->eid, req_pool_size, peer->pool_size);
+	} else {
+		// peer will likely have requested less than the available range
+		peer->pool_size = req_pool_size;
```
jk-ozlabs, line 2535: Does this mean we should also return the pool size? Or would the caller not really care?

docs/mctpd.md
```diff
@@ -203,6 +203,23 @@ This call will fail if the endpoint already has an EID, and that EID is
...
+pool-start-EID argument:
+
+```
+AssignBridgeStatic <hwaddr> <static-EID> <pool-start-EID> <pool-size>
+```
+
+`<pool-start-EID>`: The starting EID for the range of EIDs being allocated.
+This EID must be contiguous with the bridge’s own EID.
+If passed as 0, the daemon will automatically set this to `<static-EID>` + 1.
+
```
jk-ozlabs, line 222: I would suggest noting that this requirement (for the contiguous pool start) is to comply with the current 1.3.x spec. In future, the restriction ("must be contiguous […]") may be relaxed, and the behaviour of 0 implying static-EID + 1 may be changed. For the latter, I would just change that to:

> If passed as 0, mctpd will automatically choose a pool start EID

tests/test_mctpd.py
```diff
@@ -429,6 +429,88 @@ async def test_assign_endpoint_static_varies(dbus, mctpd):
...
+    pool_size = 3
+    pool_start = static_eid + 1
+
+    for _ in range(pool_size):
+        dev1.add_bridged_ep(Endpoint(iface, bytes()))
+
+    (eid, pool_start_out, _, _, new) = await mctp.call_assign_bridge_static(
+        dev1.lladdr, static_eid, pool_start, pool_size
+    )
+    print(f"dev1 assigned: eid={eid}, new={new}")
```
jk-ozlabs: No need for the print here, or below.

tests/test_mctpd.py
```diff
@@ -429,6 +429,88 @@ async def test_assign_endpoint_static_varies(dbus, mctpd):
...
+
+    dev2 = Endpoint(iface, bytes([0x1E]))
+    mctpd.network.add_endpoint(dev2)
+
+    static_eid = 12
+    pool_size = 3
+    pool_start = static_eid + 1
+
+    for _ in range(pool_size):
+        dev1.add_bridged_ep(Endpoint(iface, bytes()))
```
jk-ozlabs: `dev1` shouldn't need to be a bridge for this test; just a regular endpoint would be fine.

tests/test_mctpd.py
```diff
@@ -429,6 +429,88 @@ async def test_assign_endpoint_static_varies(dbus, mctpd):
...
+
+    assert eid == static_eid
+    assert pool_start_out == expected_pool_start
+    assert new
+
+
+async def test_assign_bridge_static_conflict(dbus, mctpd):
+    """Test that AssignBridgeStatic rejects an EID that is already in use by
+    a different endpoint.
+    """
```
jk-ozlabs, line 484: Can we also have a test that checks for a conflict in the pool range, but not the static EID?

Something like:

 dev1 -> single endpoint, EID 13
 dev2 -> bridge endpoint, EID 11, pool size 3

Adding dev2 should fail.

jk-ozlabs (changes requested):
The core code looks good, just a few comments on the doc and tests.

Also, could you rebase on top of current main? Just to make sure that we haven't created conflicts in the meantime.

jk-ozlabs (approved):
Looks great, thanks for the contribution!

PR #151, opened 2026-04-24
mctpd: fix zeroed ctrl_hdr in endpoint discovery

msnidhin:
In handle_control_endpoint_discovery(), the response header
initialization (mctp_ctrl_msg_hdr_init_resp) was placed after the
DISCOVERY_UNSUPPORTED check. When the interface has discovery
unsupported, the function would call reply_message() with a zeroed
ctrl_hdr, sending a response with no command code and no instance ID

Move the req/resp pointer setup and mctp_ctrl_msg_hdr_init_resp()
call so that response carries a correctly initialized control header.

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -1233,6 +1233,9 @@ handle_control_endpoint_discovery(struct ctx *ctx, int sd,
 		return 0;
 	}
 
+	req = (void *)buf;
+	mctp_ctrl_msg_hdr_init_resp(&resp->ctrl_hdr, *req);
+
 	if (link_data->discovered == DISCOVERY_UNSUPPORTED) {
 		resp->completion_code = MCTP_CTRL_CC_ERROR_INVALID_DATA;
 		return reply_message(ctx, sd, resp,
@@ -1243,10 +1246,6 @@ handle_control_endpoint_discovery(struct ctx *ctx, int sd,
 		// if we are already discovered (i.e, assigned an EID), then no reply
 		return 0;
 	}
-
-	req = (void *)buf;
-	resp = (void *)resp;
-	mctp_ctrl_msg_hdr_init_resp(&resp->ctrl_hdr, *req);
 
 	// we need to send using physical addressing, no entry in routing table yet
 	return reply_message_phys(ctx, sd, resp, sizeof(*resp), addr);


jk-ozlabs:
> ```
> mctpd: init response header before early returns
> 
> Fixes zeroed ctrl_hdr in Endpoint Discovery error response
> 
> Signed-off-by: Nidhin MS <nidhin.ms@intel.com>
> ```

Can we have a little explanation about the error here? For example, what behaviour are you seeing that this fixes?

src/mctpd.c
```diff
@@ -1233,6 +1233,10 @@ handle_control_endpoint_discovery(struct ctx *ctx, int sd,
 		return 0;
 	}
 
+	req = (void *)buf;
+	resp = (void *)resp;
```
mkj: This `resp` assignment (from the original) doesn't seem to do anything. May as well remove it.

msnidhin:
> > ```
> > mctpd: init response header before early returns
> > 
> > Fixes zeroed ctrl_hdr in Endpoint Discovery error response
> > 
> > Signed-off-by: Nidhin MS <nidhin.ms@intel.com>
> > ```
> 
> Can we have a little explanation about the error here? For example, what behaviour are you seeing that this fixes?

Updated PR description

src/mctpd.c
msnidhin: Hi.
In mctp_ctrl_msg_hdr_init_resp(&resp->ctrl_hdr, *req); both variables are used. 

Please confirm.

jk-ozlabs:
> > Can we have a little explanation about the error here? For example, what behaviour are you seeing that this fixes?
> 
> Updated PR description

I was more after an update to the commit message; so we have a better chance of correlating fixes to their behaviours in future.

msnidhin:
> > > Can we have a little explanation about the error here? For example, what behaviour are you seeing that this fixes?
> > 
> > 
> > Updated PR description
> 
> I was more after an update to the commit message; so we have a better chance of correlating fixes to their behaviours in future.

Hi. Updated commit message.

src/mctpd.c
jk-ozlabs: More that the statement

```
    resp = (void *)resp
```

does not do anything

jk-ozlabs (approved):
Looks good, thanks!

src/mctpd.c
msnidhin: Understood, removed the redundant line.

PR #152, opened 2026-04-29
mctpd: use same IID for retries in query_peer_properties

FreddieJheng-quanta:
Per DSP0237 Table 9, retries are retransmissions of the same MCTP control message and must reuse the same instance ID (IID). Move mctp_next_iid() from the individual query functions to the caller so one IID is shared across all retries for a given request.

Tested:
```
root@bmc:~# journalctl -u mctpd
Apr 28 22:29:09 bmc mctpd[1493]: mctpd: query_get_peer_msgtypes: Sending Get Message Type Support to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0 IID 0x00
Apr 28 22:29:09 bmc mctpd[1493]: mctpd: Retrying to get endpoint types for peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0. Attempt 1 IID 0x00
Apr 28 22:29:09 bmc mctpd[1493]: mctpd: query_get_peer_msgtypes: Sending Get Message Type Support to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0 IID 0x00
Apr 28 22:29:09 bmc mctpd[1493]: mctpd: query_get_peer_uuid: Sending Get Endpoint UUID to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0 IID 0x01
Apr 28 22:29:09 bmc mctpd[1493]: mctpd: Retrying to get peer UUID for peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0. Attempt 1 IID 0x01
Apr 28 22:29:09 bmc mctpd[1493]: mctpd: query_get_peer_uuid: Sending Get Endpoint UUID to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0 IID 0x01
```

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -2540,20 +2540,21 @@ static int get_endpoint_peer(struct ctx *ctx, sd_bus_error *berr,
 	return 0;
 }
 
-static int query_get_peer_msgtypes(struct peer *peer)
+static int query_get_peer_msgtypes(struct peer *peer, uint8_t iid)
 {
 	struct sockaddr_mctp_ext addr;
 	struct mctp_ctrl_cmd_get_msg_type_support req;
 	struct mctp_ctrl_resp_get_msg_type_support *resp = NULL;
 	uint8_t *buf = NULL;
 	size_t buf_size, expect_size;
-	uint8_t iid;
 	int rc;
 
 	peer->num_message_types = 0;
 	free(peer->message_types);
 	peer->message_types = NULL;
-	iid = mctp_next_iid(peer->ctx);
+
+	warnx("%s: Sending Get Message Type Support to %s IID 0x%02x", __func__,
+	      peer_tostr(peer), iid);
 
 	mctp_ctrl_msg_hdr_init_req(&req.ctrl_hdr, iid,
 				   MCTP_CTRL_CMD_GET_MESSAGE_TYPE_SUPPORT);
@@ -2744,14 +2745,13 @@ out:
 	return rc;
 }
 
-static int query_get_peer_uuid(struct peer *peer)
+static int query_get_peer_uuid(struct peer *peer, uint8_t iid)
 {
 	struct sockaddr_mctp_ext addr;
 	struct mctp_ctrl_cmd_get_uuid req;
 	struct mctp_ctrl_resp_get_uuid *resp = NULL;
 	uint8_t *buf = NULL;
 	size_t buf_size;
-	uint8_t iid;
 	int rc;
 
 	if (peer->state != REMOTE) {
@@ -2760,7 +2760,8 @@ static int query_get_peer_uuid(struct peer *peer)
 		return -EPROTO;
 	}
 
-	iid = mctp_next_iid(peer->ctx);
+	warnx("%s: Sending Get Endpoint UUID to %s IID 0x%02x", __func__,
+	      peer_tostr(peer), iid);
 
 	mctp_ctrl_msg_hdr_init_req(&req.ctrl_hdr, iid,
 				   MCTP_CTRL_CMD_GET_ENDPOINT_UUID);
@@ -3133,20 +3134,26 @@ static int query_peer_properties(struct peer *peer)
 {
 	const unsigned int max_retries = 4;
 	bool supports_vdm = false;
+	uint8_t iid;
 	int rc;
 
+	iid = mctp_next_iid(peer->ctx);
 	for (unsigned int i = 0; i < max_retries; i++) {
-		rc = query_get_peer_msgtypes(peer);
+		rc = query_get_peer_msgtypes(peer, iid);
 
 		// Success
 		if (rc == 0)
 			break;
 
-		// On timeout, retry
+		// On timeout, retry with same IID
 		if (rc == -ETIMEDOUT) {
 			if (peer->ctx->verbose)
-				warnx("Retrying to get endpoint types for %s. Attempt %u",
-				      peer_tostr(peer), i + 1);
+				warnx("Retrying to get endpoint types for %s. Attempt %u IID 0x%02x",
+				      peer_tostr(peer), i + 1, iid);
+			if (i + 1 == max_retries)
+				warnx("Get Message Type Support timed out for %s, "
+				      "proceeding to Get UUID without response",
+				      peer_tostr(peer));
 			rc = 0;
 			continue;
 		}
@@ -3194,18 +3201,22 @@ static int query_peer_properties(struct peer *peer)
 		}
 	}
 
+	iid = mctp_next_iid(peer->ctx);
 	for (unsigned int i = 0; i < max_retries; i++) {
-		rc = query_get_peer_uuid(peer);
+		rc = query_get_peer_uuid(peer, iid);
 
 		// Success
 		if (rc == 0)
 			break;
 
-		// On timeout, retry
+		// On timeout, retry with same IID
 		if (rc == -ETIMEDOUT) {
 			if (peer->ctx->verbose)
-				warnx("Retrying to get peer UUID for %s. Attempt %u",
-				      peer_tostr(peer), i + 1);
+				warnx("Retrying to get peer UUID for %s. Attempt %u IID 0x%02x",
+				      peer_tostr(peer), i + 1, iid);
+			if (i + 1 == max_retries)
+				warnx("Get Endpoint UUID timed out for %s",
+				      peer_tostr(peer));
 			rc = 0;
 			continue;
 		}
--- a/tests/test_mctpd.py
+++ b/tests/test_mctpd.py
@@ -1765,6 +1765,54 @@ async def test_query_peer_properties_retry_timeout(nursery, dbus, sysnet):
     assert res == 0
 
 
+async def test_query_peer_properties_same_iid_on_retry(nursery, dbus, sysnet):
+    """Verify that retries for query_peer_properties reuse the same IID.
+
+    Per DSP0237 Table 9, a retry is a retransmission of the same MCTP control
+    message and must use the same instance ID (IID) within MT4.
+    """
+
+    class IIDTrackingEndpoint(Endpoint):
+        """Drop the first Get Message Type Support request, record all IIDs seen."""
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.seen_iids = []
+            self.drop_next = True
+
+        async def handle_mctp_control(self, sock, addr, data):
+            rq = data[0] & 0x80
+            opcode = data[1]
+            iid = data[0] & 0x1F
+            if rq and opcode == 0x05:  # Get Message Type Support
+                self.seen_iids.append(iid)
+                if self.drop_next:
+                    self.drop_next = False
+                    return  # simulate timeout
+            return await super().handle_mctp_control(sock, addr, data)
+
+    mctpd = MctpdWrapper(dbus, sysnet)
+    await mctpd.start_mctpd(nursery)
+
+    iface = mctpd.system.interfaces[0]
+    mctp = await mctpd_mctp_iface_obj(dbus, iface)
+
+    ep = IIDTrackingEndpoint(iface, bytes([0x1A]), eid=15, types=[0, 1, 2])
+    mctpd.network.add_endpoint(ep)
+
+    await mctp.call_setup_endpoint(ep.lladdr)
+
+    # Two Get Message Type Support requests: initial + one retry
+    assert len(ep.seen_iids) == 2
+    # Both must carry the same IID
+    assert ep.seen_iids[0] == ep.seen_iids[1], (
+        f"IID changed across retry: {ep.seen_iids[0]:#04x} -> {ep.seen_iids[1]:#04x}"
+    )
+
+    res = await mctpd.stop_mctpd()
+    assert res == 0
+
+
 async def test_bridged_endpoint_poll(dbus, sysnet, nursery, autojump_clock):
     """Test that we use endpoint poll interval from the config and
     that we discover bridged endpoints via polling


jk-ozlabs:
> Per DSP0237 Table 9

That table just covers the timeouts for the SMBus binding, and doesn't seem to cover any IID reuse behaviour of the control protocol. Can you elaborate on what's requiring this?

I would prefer separate IIDs for retries, as that allows us to track delays vs. lost messages.

FreddieJheng-quanta:
> > Per DSP0237 Table 9
> 
> That table just covers the timeouts for the SMBus binding, and doesn't seem to cover any IID reuse behaviour of the control protocol. Can you elaborate on what's requiring this?
> 
> I would prefer separate IIDs for retries, as that allows us to track delays vs. lost messages.

Hi @jk-ozlabs, thanks for the review!
Wanted to discuss further the reasoning behind this implementation.

We encountered a PLDM device that was slow to respond during init sync, causing the requester to time out and switch to a new IID. The vendor's position was that the BMC should retry with the same IID within the MT4 window (5~6 sec) rather than treating it as a new request.

Before implementing this, I came across DSP0236 Table 10, which describes the Instance ID field as:
_"used to identify new instances of an MCTP control Request or Datagram to differentiate new requests or datagrams that are sent to a given message terminus from retried messages that are sent to the same message terminus."_

One reading is that the IID value itself signals the distinction: same IID = retry, different IID = new request, which would imply retries should preserve the IID. Though the spec doesn't explicitly say "retries must reuse the same IID" so it seems open to interpretation.

Would be great to hear your take on this!

jk-ozlabs:
OK, it sounds like we would be justified in using the same IID within the retry loop - I would suggest updating the commit to reference DSP0236 Table 10 though.

This is probably the point at which we need to converge the multiple separate retry loops though, so we have consistent behaviour. Is that something you want to look at doing beforehand?

jk-ozlabs:
(I might take a look at coalescing those retry loops now)

FreddieJheng-quanta:
> OK, it sounds like we would be justified in using the same IID within the retry loop - I would suggest updating the commit to reference DSP0236 Table 10 though.
> 
> This is probably the point at which we need to converge the multiple separate retry loops though, so we have consistent behaviour. Is that something you want to look at doing beforehand?

Thanks, Will update the commit to reference DSP0236 Table 10. Regarding coalescing the retry loops, note that `query_get_peer_vdm_types` doesn't take an iid parameter currently.

jk-ozlabs:
OK, I have refactored the various retry loops into one, and provided a facility to disable retries. In doing that, the IIDs are naturally repeated for command retries.

I have then taken the test case from your commit (preserving authorship for that), which passes.

This is in my `experimental/refactor-retry` branch: https://github.com/jk-ozlabs/mctp/tree/experimental/refactor-retry . Let me know what you think. If that's sensible, I'll propose for an actual PR.

I will likely have to modify some of the retry semantics, we probably don't want to retry on the downstream bridge pool probes either.

FreddieJheng-quanta:
> OK, I have refactored the various retry loops into one, and provided a facility to disable retries. In doing that, the IIDs are naturally repeated for command retries.
> 
> I have then taken the test case from your commit (preserving authorship for that), which passes.
> 
> This is in my `experimental/refactor-retry` branch: https://github.com/jk-ozlabs/mctp/tree/experimental/refactor-retry . Let me know what you think. If that's sensible, I'll propose for an actual PR.
> 
> I will likely have to modify some of the retry semantics, we probably don't want to retry on the downstream bridge pool probes either.

Thanks for the refactor, It looks much cleaner, though I haven't run the tests yet.

One minor correction: the docstring in `test_query_peer_properties_same_iid_on_retry` carried over my incorrect reference to DSP0237 Table 9, it should be DSP0236 Table 10.

Sorry that I’ll be away next week and won't be able to reply until I’m back on May 11th. Thanks for understanding!

jk-ozlabs:
> It looks much cleaner, though I haven't run the tests yet.

Do you mean the in-tree tests? or do you have your own test setup?

>  it should be DSP0236 Table 10.

ack, I shall update.

I'll keep the branch around, no rush on that. If you can let me know how this goes for testing, I'll hold off on merging until then.

FreddieJheng-quanta:
> > It looks much cleaner, though I haven't run the tests yet.
> 
> Do you mean the in-tree tests? or do you have your own test setup?
> 
> > it should be DSP0236 Table 10.
> 
> ack, I shall update.
> 
> I'll keep the branch around, no rush on that. If you can let me know how this goes for testing, I'll hold off on merging until then.

I noticed that commit `6cce228` inadvertently made UUID errors fatal. I've tested on my environment and prepared a fix on top of your branch: https://github.com/FreddieJheng-quanta/mctp/tree/fix/uuid-nonfatal-refactor-retry

jk-ozlabs:
> I noticed that commit `6cce228` inadvertently made UUID errors fatal. I've tested on my environment and prepared a fix on top of your branch: https://github.com/FreddieJheng-quanta/mctp/tree/fix/uuid-nonfatal-refactor-retry

Good catch, thanks. Even if it matches previous behaviour, I don't think the fix is 100% correct though - we want to continue on all errors from querying the UUID (and any other optional commands). I'll have a look, and I think we can simplify that a litte

FreddieJheng-quanta:
> > I noticed that commit `6cce228` inadvertently made UUID errors fatal. I've tested on my environment and prepared a fix on top of your branch: https://github.com/FreddieJheng-quanta/mctp/tree/fix/uuid-nonfatal-refactor-retry
> 
> Good catch, thanks. Even if it matches previous behaviour, I don't think the fix is 100% correct though - we want to continue on all errors from querying the UUID (and any other optional commands). I'll have a look, and I think we can simplify that a litte

Thanks, that makes sense. Handling these optional property queries as non-fatal is a better approach. Looking forward to it.

jk-ozlabs:
OK, I have updated that branch, and will convert to a new PR. Let me know if you have any objections.

jk-ozlabs:
New PR up at https://github.com/CodeConstruct/mctp/pull/154

FreddieJheng-quanta:
> OK, I have updated that branch, and will convert to a new PR. Let me know if you have any objections.

I hit a Wrong IID error on `query_get_peer_msgtypes` in our environment:
```
query_get_peer_msgtypes: Sending Get Message Type Support to peer eid 50 ... IID 0x00
read_message got from sockaddr_mctp_ext eid 50 ... len 6
[peer 1:50, cmd Get Message Type Support]: Wrong IID (0x1f, expected 0x00)
Error getting endpoint types for peer ...: No message of desired type
emitting endpoint add: .../endpoints/50

```

My understanding is that `endpoint_query_addr` opens a new socket per query. Since MCTP sockets are type-based, the kernel delivers any arriving control message to the currently open socket regardless of which query sent the request. When the previous round's response arrives late (after `close(sd)`), it gets delivered to the next query's socket.

Does this match your understanding? A possible fix would be to reuse a single socket across all sub-queries in query_peer_properties to avoid stale responses crossing query boundaries.

jk-ozlabs:
> Since MCTP sockets are type-based, the kernel delivers any arriving control message to the currently open socket regardless of which query sent the request.

We're not `bind()`-ing on these sockets though, so the only incoming messages we should see will be replies to prior messages we have sent, determined entirely by the tag value on the incoming responses (and not by message type).

It may be that we are re-using the tags too fast though - so a new socket consumes the tag that was dropped by an old socket's close(), so the tag on the (old) incoming reply is now valid, and matches the new socket. 

Given that you're seeing different IIDs, this was a response from a different command, not a retry.

How frequently are commands being issued in your test case?

jk-ozlabs:
... but yes, using a single socket would address this; the tags used on re-sent messages would remain valid for the entire tag timeout, so the responses on messages that we have timed-out (ie, have an invalid IID) would still need to be ignored.

jk-ozlabs:
actually, we can probably just do the IID check earlier, and resume the recvmsg() on failure, as part of the retry mechanism.

FreddieJheng-quanta:
> > Since MCTP sockets are type-based, the kernel delivers any arriving control message to the currently open socket regardless of which query sent the request.
> 
> We're not `bind()`-ing on these sockets though, so the only incoming messages we should see will be replies to prior messages we have sent, determined entirely by the tag value on the incoming responses (and not by message type).
> 
> It may be that we are re-using the tags too fast though - so a new socket consumes the tag that was dropped by an old socket's close(), so the tag on the (old) incoming reply is now valid, and matches the new socket.
> 
> Given that you're seeing different IIDs, this was a response from a different command, not a retry.
> 
> How frequently are commands being issued in your test case?

Thanks for the correction. 

Here are the full logs. 

```
# Round 1: Set Endpoint ID — timeout x4
May 12 21:37:44 bmc mctpd[1479]: mctpd: method_assign_endpoint: called by :1.185
May 12 21:37:44 bmc mctpd[1479]: add_peer_neigh: Adding neigh to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0
May 12 21:37:44 bmc mctpd[1479]: add_peer_route: Adding route to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0
May 12 21:37:44 bmc mctpd[1479]: mctpd: receive timed out from sockaddr_mctp_ext eid 0 net 0 type 0x00 if 11 hw len 1 0x62, attempt 1/4
May 12 21:37:45 bmc mctpd[1479]: mctpd: receive timed out from sockaddr_mctp_ext eid 0 net 0 type 0x00 if 11 hw len 1 0x62, attempt 2/4
May 12 21:37:45 bmc mctpd[1479]: mctpd: receive timed out from sockaddr_mctp_ext eid 0 net 0 type 0x00 if 11 hw len 1 0x62, attempt 3/4
May 12 21:37:46 bmc mctpd[1479]: mctpd: receive timed out from sockaddr_mctp_ext eid 0 net 0 type 0x00 if 11 hw len 1 0x62, attempt 4/4
May 12 21:37:46 bmc mctpd[1479]: unpublish_peer: Deleting neigh to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0
May 12 21:37:46 bmc mctpd[1479]: unpublish_peer: Deleting route to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0
May 12 21:37:46 bmc mctpd[1479]: mctpd: Returning dbus error 'MCTP Endpoint did not respond', errcode=-110 (Connection timed out)

# Round 2: host retries 3s later
May 12 21:37:49 bmc mctpd[1479]: mctpd: method_assign_endpoint: called by :1.185
May 12 21:37:49 bmc mctpd[1479]: add_peer_neigh: Adding neigh to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0
May 12 21:37:49 bmc mctpd[1479]: add_peer_route: Adding route to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0
May 12 21:37:49 bmc mctpd[1479]: mctpd: receive timed out from sockaddr_mctp_ext eid 0 net 0 type 0x00 if 11 hw len 1 0x62, attempt 1/4
May 12 21:37:49 bmc mctpd[1479]: mctpd: read_message got from sockaddr_mctp_ext eid 50 net 1 type 0x00 if 11 hw len 1 0x62 len 6
May 12 21:37:49 bmc mctpd[1479]: add_peer_neigh: Adding neigh to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0
May 12 21:37:49 bmc mctpd[1479]: add_peer_route: Adding route to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0

# query_peer_properties
May 12 21:37:49 bmc mctpd[1479]: mctpd: query_get_peer_msgtypes: Sending Get Message Type Support to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0 IID 0x00
May 12 21:37:50 bmc mctpd[1479]: mctpd: read_message got from sockaddr_mctp_ext eid 50 net 1 type 0x00 if 11 hw len 1 0x62 len 6
May 12 21:37:50 bmc mctpd[1479]: mctpd: [peer 1:50, cmd Get Message Type Support]: Wrong IID (0x1f, expected 0x00)
May 12 21:37:50 bmc mctpd[1479]: mctpd: Error getting endpoint types for peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0: No message of desired type
May 12 21:37:50 bmc mctpd[1479]: mctpd: query_get_peer_uuid: Sending Get Endpoint UUID to peer eid 50 net 1 phys physaddr if 11 hw len 1 0x62 state 0 IID 0x01
May 12 21:37:50 bmc mctpd[1479]: mctpd: receive timed out from sockaddr_mctp_ext eid 50 net 1 type 0x00 if 0 hw len 0 0x, attempt 1/4
May 12 21:37:50 bmc mctpd[1479]: mctpd: read_message got from sockaddr_mctp_ext eid 50 net 1 type 0x00 if 11 hw len 1 0x62 len 19
May 12 21:37:50 bmc mctpd[1479]: mctpd: emitting endpoint add: /au/com/codeconstruct/mctp1/networks/1/endpoints/50
```

From the logs, does this look like it matches your suspicion?

jk-ozlabs:
Thanks. Looks like it, yes.

Given you're seeing the wrong IID on the Get Message Type support query, and it's the immediately-prior IID, the invalid response must have been from the Set Endpoint ID that timed-out earlier (I assume this isn't a bridge that you're probing). This means that using the single socket for the query_peer_properties wouldn't be sufficient, as the invalid IID was from outside that "session" of commands.

I think the IID handling improvement would be a valid fix for this. We'd have the same issue regardless of the retry-refactor change, so I'll propose that fix separately.

FreddieJheng-quanta:
> Thanks. Looks like it, yes.
> 
> Given you're seeing the wrong IID on the Get Message Type support query, and it's the immediately-prior IID, the invalid response must have been from the Set Endpoint ID that timed-out earlier (I assume this isn't a bridge that you're probing). This means that using the single socket for the query_peer_properties wouldn't be sufficient, as the invalid IID was from outside that "session" of commands.
> 

EID 50 is a direct endpoint, configured via Entity Manager with an `MCTPI2CTarget` configuration.
The len 6 in the log also matches `mctp_ctrl_resp_set_eid` size, further confirming the packet was a stale Set Endpoint ID response.

> I think the IID handling improvement would be a valid fix for this. We'd have the same issue regardless of the retry-refactor change, so I'll propose that fix separately.

got it, thanks

jk-ozlabs:
#154 has been merged now; I don't think we need this PR at this point, so closing. Feel free to reopen with a comment if we need to revisit.

PR #154, opened 2026-05-13
Command submission / completion refactor, making retry logic common

jk-ozlabs:
A bit of a refactor to the MCTP command submission path, which allows us to implement a common retry mechanism (rather than having each top-level command caller implement its own retry logic).

Commands are now represented via a new struct:

```c
struct mctp_ctrl_cmd {
	/* populated by caller */
	void *req;
	size_t req_len;
	bool disable_retry;

	/* populated on response RX */
	void *resp;
	size_t resp_len;
	struct sockaddr_mctp_ext resp_addr;
};
```

\- which encapsulates the request & response buf/length, and retry behaviour.

This PR is based on functionality requested in https://github.com/CodeConstruct/mctp/pull/152; we want to re-use the command IID for retries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
    now called `role`. Configuration parsing will still allow the `mode` setting,
    but this will be deprecated in a later release.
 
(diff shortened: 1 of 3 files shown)


FreddieJheng-quanta:
Thanks for the refactor

PR #157, opened 2026-06-11
mctpd: Improve sequencing of Set Endpoint ID responses in endpoint mode

### LINKED ISSUE #155: SetEID response time
@jk-ozlabs 

In [SetEID](https://github.com/CodeConstruct/mctp/blob/4186b5af06f66447a5b58290277ecfa3b7f182aa/src/mctpd.c#L826) handler the response is sent only after call to [setup_peer](https://github.com/CodeConstruct/mctp/blob/4186b5af06f66447a5b58290277ecfa3b7f182aa/src/mctpd.c#L902) . 

This will add additional delay in sending SetEID response back to bus owner which can result in timeout also. 

Is it good to send the SetEID response to BO and then setup the peer properties.

jk-ozlabs:
When running in endpoint mode, mctpd's Set Endpoint ID command handler will trigger an enumeration of the peer (ie, the bus owner that issued the Set Endpoint ID command).

Currently, that enumeration occurs before we send the Set Endpoint ID response, which is a little unexpected: at that point, the bus owner has not received any indication that the command completed successfully. Also, the wait on the set of peer-enumeration commands may delay the Set Endpoint ID response beyond a command timeout.

So, respond to the Set Endpoint ID command first, then perform peer enumeration. This requires a minor rework to the `setup_endpoint_peer()` process, as we will need to establish routes beforehand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
    re-use outgoing control protocol instance IDs (IIDs) when we re-send
    commands. More instances of commands now have retry logic too.
 
+3. When running in endpoint mode, `mctpd` will now issue responses to incoming
+   Set Endpoint ID commands before performing enumeration on the (bus-owner)
+   peer.
+
 ## [2.5] - 2026-02-17
 
 ### Added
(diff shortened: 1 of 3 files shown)


jk-ozlabs:
@khangng-ampere let me know if you have any input, given you're one of the main users of endpoint mode.

msnidhin:
Hi @jk-ozlabs  
Can we get this merged?  This change is fixing issue #155

khangng-ampere:
> @khangng-ampere let me know if you have any input, given you're one of the main users of endpoint mode.

Looks good to me! Thanks for fixing it

PR #159, opened 2026-06-29
mctpd: Automatic Discovery Notify for endpoint role

msnidhin:
Some physical transports (PCIe VDM and I3C) require an MCTP endpoint to send
a Discovery Notify control message to the bus owner before the bus owner will
enumerate it. Add support to mctpd for sending this automatically, without
requiring an external caller to drive it.

Introduce a SetBusOwner method on the au.com.codeconstruct.MCTP.Endpoint1
D-Bus interface (available on per-interface objects when the link role is
Endpoint). This allows an external management application to supply the
physical address of the bus owner when mctpd cannot determine it from the
transport binding itself.

### DIFF
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -245,6 +245,48 @@ reports as a bridge.
 
 Bridge endpoints should be initialised with `AssignEndpoint` instead.
 
+### Endpoint interface: `au.com.codeconstruct.MCTP.Endpoint1` interface
+
+When the interface `Role` is `Endpoint`, the MCTP interface object also hosts
+the `au.com.codeconstruct.MCTP.Endpoint1` D-Bus interface. This exposes
+endpoint-role functions on the per-interface D-Bus path.
+
+```
+NAME                                  TYPE      SIGNATURE RESULT/VALUE FLAGS
+au.com.codeconstruct.MCTP.Interface1  interface -         -            -
+.NetworkId                            property  u         1            emits-change
+.Role                                 property  s         "Endpoint"   emits-change writable
+au.com.codeconstruct.MCTP.Endpoint1   interface -         -            -
+.BusOwner                             property  ay        -            writable
+```
+
+#### `.BusOwner`: `ay`
+
+Sets the physical address of the bus owner for this interface. For transport
+types that know the bus owner physaddr (or do not need one to address the bus
+owner), this would be populated automatically by the transport binding. For
+other cases, this property allows the address to be supplied by an external
+management application.
+
+On transport types that require a Discovery Notify (PCIe and I3C), setting the
+bus owner address will automatically trigger a Discovery Notify to the bus
+owner if the interface is in Endpoint role and not yet discovered.
+
+`BusOwner <hwaddr>`
+
+ - `<hwaddr>` Physical hardware address of the bus owner. For MCTP-over-I3C
+   this is the 6-byte Provisional ID (PID) of the bus owner.
+
+An example for MCTP-over-I3C, setting the bus owner's 6-byte PID
+`00 6c 90 01 23 45`:
+
+```shell
+busctl set-property au.com.codeconstruct.MCTP1 \
+    /au/com/codeconstruct/mctp1/interfaces/mctpi3c0 \
+    au.com.codeconstruct.MCTP.Endpoint1 \
+    BusOwner ay 6 0x00 0x6c 0x90 0x01 0x23 0x45
+```
+
 ## Network objects: `/au/com/codeconstruct/mctp1/networks/<net>`
 
 These objects represent MCTP networks which have been added use `mctp link`
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -151,7 +151,11 @@ struct link {
 	char *path;
 	sd_bus_slot *slot_iface;
 	sd_bus_slot *slot_busowner;
+	sd_bus_slot *slot_endpoint;
 	sd_event_source *role_defer;
+
+	/* Physical address of the bus owner. */
+	dest_phys bus_owner;
 
 	struct ctx *ctx;
 };
@@ -389,6 +393,7 @@ static const sd_bus_vtable bus_endpoint_obmc_vtable[];
 static const sd_bus_vtable bus_endpoint_cc_vtable[];
 static const sd_bus_vtable bus_endpoint_bridge[];
 static const sd_bus_vtable bus_endpoint_uuid_vtable[];
+static const sd_bus_vtable bus_link_endpoint_vtable[];
 
 __attribute__((format(printf, 1, 2))) static void bug_warn(const char *fmt, ...)
 {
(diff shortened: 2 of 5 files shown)


msnidhin:
Follow up PR for 
https://discord.com/channels/775381525260664832/778790638563885086/1488825384471625799

docs/mctpd.md
```diff
@@ -248,6 +218,43 @@ reports as a bridge.
...
+au.com.codeconstruct.MCTP.Endpoint1   interface -         -            -
+.AttemptDiscoveryNotify               method    ay        -            -
+```
+
+#### `.AttemptDiscoveryNotify`: `ay`
+
+Some physical transports (such as MCTP-over-I3C) require the endpoint to
+send a Discovery Notify to announce its presence before the bus owner will
+enumerate it. On those interfaces this method can be called before the
+endpoint will be assigned an EID.
```
jk-ozlabs: This isn't exactly correct; in most cases, the I3C peer has explicit knowledge of the presence of the new MCTP endpoint.

I'm not clear on your specific use-case here, could you expand this on the scenario where this is required?

docs/mctpd.md
```diff
@@ -248,6 +218,43 @@ reports as a bridge.
...
+```
+NAME                                  TYPE      SIGNATURE RESULT/VALUE FLAGS
+au.com.codeconstruct.MCTP.Interface1  interface -         -            -
+.NetworkId                            property  u         1            emits-change
+.Role                                 property  s         "Endpoint"   emits-change writable
+au.com.codeconstruct.MCTP.Endpoint1   interface -         -            -
+.AttemptDiscoveryNotify               method    ay        -            -
+```
+
+#### `.AttemptDiscoveryNotify`: `ay`
```
jk-ozlabs: Since this is only applicable for certain transport binding types, we may want to put this into an interface that is specific to those transports, rather than `Endpoint1` which should be common across all. Any thoughts on that?

src/mctpd.c
```diff
@@ -1675,7 +1660,7 @@ static int mctp_ctrl_validate_response(struct mctp_ctrl_cmd *cmd,
 {
 	struct mctp_ctrl_resp *rsp;
 
-	if (exp_size <= sizeof(*rsp)) {
+	if (exp_size < sizeof(*rsp)) {
```
jk-ozlabs, line 1687: I'd prefer this as a separate fix.

src/mctpd.c
```diff
@@ -4246,6 +4019,67 @@ static int method_register_vdm_type_support(sd_bus_message *call, void *data,
...
+					   sd_bus_error *berr)
+{
+	struct mctp_ctrl_cmd_discovery_notify req = { 0 };
+	struct mctp_ctrl_resp_discovery_notify *resp = NULL;
+	dest_phys desti = { 0 }, *dest = &desti;
+	struct link *link = data;
+	struct ctx *ctx = link->ctx;
+	struct mctp_ctrl_cmd cmd = { 0 };
+	uint8_t iid;
+	int rc;
```
jk-ozlabs: Minor, but if you're reworking, can you reverse-christmas-tree these?

src/mctpd.c
```diff
@@ -4246,6 +4019,67 @@ static int method_register_vdm_type_support(sd_bus_message *call, void *data,
...
+
+	dest->ifindex = link->ifindex;
+	if (dest->ifindex <= 0)
+		return sd_bus_error_setf(berr, SD_BUS_ERROR_INVALID_ARGS,
+					 "Unknown MCTP interface");
+
+	rc = message_read_hwaddr(call, dest);
+	if (rc < 0)
+		goto err;
```
jk-ozlabs: This is taking the `mctp_ctrl_cmd_free` path, but the surrounding conditionals do not. You don't have `cmd->resp` set here, so I don't think you need the cleanup.

src/mctpd.c
```diff
@@ -4246,6 +4019,67 @@ static int method_register_vdm_type_support(sd_bus_message *call, void *data,
...
+					 dest_phys_tostr(dest), iid,
+					 MCTP_CTRL_CMD_DISCOVERY_NOTIFY);
+	if (rc)
+		goto err;
+
+	mctp_ctrl_cmd_free(&cmd);
+	return sd_bus_reply_method_return(call, "");
+err:
+	mctp_ctrl_cmd_free(&cmd);
+	set_berr(ctx, rc, berr);
```
jk-ozlabs, line 4378: What's the significance of returning an error in the API here? Do we need to fail this at all?

docs/mctpd.md
```diff
@@ -248,6 +218,43 @@ reports as a bridge.
 
 Bridge endpoints should be initialised with `AssignEndpoint` instead.
 
+### Endpoint interface: `au.com.codeconstruct.MCTP.Endpoint1` interface (on interface objects)
```
jk-ozlabs: drop the "(on interface objects)"; this is covered by the "Endpoint Interface" term there (it's a bit confusing on the two meanings of "interface" here, but I don't think a third reference helps).

jk-ozlabs (changes requested):
The implementation looks good here (some fairly minor comments inline) but there are some items that I don't think we had covered on the earlier discussion.

As a naïve approach, we could do the Discovery Notify automatically when we're bringing-up an interface that is one of the Discovery-Notify-requiring types. However, that's not always possible, because:

 - we don't always know when that is required (as DSP0233 specifies)
 - we don't always know how to address the bus owner

Could you elaborate on your use-case here to see how we should be handling the former? If it's just a matter of knowing the bus-owner address, this could possibly be designed as more of an interface to supply the bus-owner address, and then we would decide on sending the Discovery Notify as appropriate. If that's not feasible, that's fine, but I would like to know the factors involved there.

src/mctpd.c
jk-ozlabs: (to be clear: yes, we should fail on invalid arguments etc, but do we need to fail on specifics of the peer response - or lack of one?)

docs/mctpd.md
msnidhin: Hi. In my use case i3c address assignment happened early in the boot. At that time the mctp stack in endpoint mode was not ready. It took some time for mctpd service to start. After that I want to send discovery notify and announce endpoint presence.

docs/mctpd.md
msnidhin: OK. Like only if link->phys_binding == MCTP_PHYS_BINDING_PCIE_VDM or link->phys_binding == MCTP_PHYS_BINDING_I3C ?

also what interface name to be used ?

docs/mctpd.md
jk-ozlabs: OK, so it would work to have mctpd automatically sent the Dicovery Notify when the local interface is found? (or when the role is set to Endpoint?)

That way you don't need the dbus API at all, nor external infrastructure to call it.

Or does this need to be sequenced with other things?

msnidhin: Hi. sending the command when Role is set to endpoint also works. 
But as discussed before mctpd won't know the physical address of the BO. Some external apps like mctpreactor should supply this.

src/mctpd.c
msnidhin: Hi. Intention was to retry on failure. Like if BO sends ERR_NOT_READY CC or it is not ready to accept due to some other reasons.

src/mctpd.c
msnidhin: PR added. Keeping it here to prevent UT failure

src/mctpd.c
msnidhin: Sure

src/mctpd.c
msnidhin: returned rc directly

docs/mctpd.md
msnidhin: Done

docs/mctpd.md
jk-ozlabs: I think that as an endpoint, we should be able to address the BO without a-prior knowledge of any physical addresses. The IBI mechanism for I3C should mean that no phys addresses are required for target-to-primary commuinication, right?

Given there is no upstream I3C target implementation though, we don't have any specifics of that.

src/mctpd.c
jk-ozlabs: That's requiring that caller to implement too much of the control protocol semantics. Errors / retrying should be handled by mctpd.

docs/mctpd.md
msnidhin: Hi. The use case is the red network in this image from DSP0233. 
Section 5.1.3 MCTP Bus Owner for I3C bus 
image
MCTP BO is an I3C target. I know the PID of the BO and need to send Discovery Notify. 
And I didn't find any method to send this command without using the address of the BO.

jk-ozlabs: My point is that the endpoint must already know the phys address of the bus owner in that case (because, as the I3C controller, it assigned the I3C address to it!)

But since I have no idea what your MCTP-over-I3C transport binding implementation looks like, I can't help with how you could hook that into mctpd. I would suggest that mctpd could handle this entirely internally.

Please propose your binding implementation upstream, so we can reason on a good solution for this.

jk-ozlabs: Sorry, my confusion there - it had occurred to me that you _are_ using the upstream MCTP-over-I3C-(controller) binding, just in MCTP endpoint mode (ie., implementing the Bridge 2 in your diagram).

Your unique case here is that the transport itself cannot provide the bus owner physaddr, as it could be any one of the I3C targets connected to the controller.

In which case, I think a "set the bus owner" interface would be more appropriate, and would be applicable to the `Endpoint1` interface in general. For transport types that know the bus owner physaddr (or do not need one to address the BO), this would be populated by the transport binding (typically, via netlink link config). For your case, it would be through the new interface to set the bus owner.

The flow would be:

When all of the following are true:
  - the interface has been discovered
  - the interface role is set as an endpoint
  - the interface type allows discovery notify
  - we know the bus owner physaddr

Then we start the Discovery Notify process:
 - send a discovery notify to the BO
 - on timeout (or non-fatal errors) retry after some reasonable delay
 - on actual discovery, cease the Discovery notify process

That then solves the below issue around having the dbus caller have to handle messaging errors; setting the bus owner may start the discovery process, rather than having the caller be responsible for making progress.

I'd need to confirm this fits with the full discovery process required for other binding specs, but that's a job for Monday.

jk-ozlabs:
Thanks for the updates.

Your new set still has the `AttemptDiscoveryNotify` dbus call - do we still need that though? mctpd should be able to work out when it needs discovery automatically, no?

msnidhin:
> Thanks for the updates.
> 
> Your new set still has the `AttemptDiscoveryNotify` dbus call - do we still need that though? mctpd should be able to work out when it needs discovery automatically, no?

Hi Sure. I will push the changes to align with the comment.
> When all of the following are true:
> 
> * the interface has been discovered
> * the interface role is set as an endpoint
> * the interface type allows discovery notify
> * we know the bus owner physaddr
> 
> Then we start the Discovery Notify process:
> 
> * send a discovery notify to the BO
> * on timeout (or non-fatal errors) retry after some reasonable delay
> * on actual discovery, cease the Discovery notify process

docs/mctpd.md
```diff
@@ -248,6 +248,48 @@ reports as a bridge.
...
+.Role                                 property  s         "Endpoint"   emits-change writable
+au.com.codeconstruct.MCTP.Endpoint1   interface -         -            -
+.SetBusOwner                          method    ay        -            -
+```
+
+#### `.SetBusOwner`: `ay`
+
+Sets the physical address of the bus owner for this interface. For transport
+types that know the bus owner physaddr (or do not need one to address the bus
+owner), this would be populated automatically by the transport binding. For
```
jk-ozlabs, line 267: This sounds like it should be a (writable) property then? Should clients be able to query the current bus owner value?

msnidhin: Changed to property. 
I think getter is needed for non-basic type properties

jk-ozlabs: I think that looks good for the interface.

And yes, all good with the explicit get & set methods.

jk-ozlabs: One thought here: what's the API intention for when no bus owner is set? and how do we distinguish that from cases where there is no physical addressing mechanism (eg, USB & serial transports).

We should probably document that here too.

msnidhin: Hi. I see only I3C and PCIe spec mention abount sending discovery notify to announce endpoint presence. 

can I add to doc like this ?
For transport types that do not use physical addressing for bus owner communication (such as USB and serial), this property has no effect; Discovery Notify is only triggered on transports that require it (PCIe VDM and I3C).

msnidhin: Updated docs..

PR #160, opened 2026-07-02
mctpd: fix mctp_ctrl_validate_response size check

msnidhin:
fix mctp_ctrl_validate_response to use strict less-than when comparing exp_size against sizeof(struct mctp_ctrl_resp). The previous <= incorrectly rejected responses whose complete structure is exactly the base response size, such as Discovery Notify.


jk-ozlabs:
Thanks!

PR #162, opened 2026-07-03
mctpd: Require mandatory peer query commands only

### LINKED ISSUE #161: Query on AssignEndpoint method
Hi @jk-ozlabs  . 
During testing I observed that AssignEndpoint method return failure if the endpoint responds with nonsuccess CC for GetUUID command. 
Is it expected? Since GetUUID is not a mandatory command.

jk-ozlabs:
We don't want to fail peer enumeration if the optional Get Endpoint UUID and Get Vendor Defined Message Support commands are unimplemented, so return success from query_peer_properties if these fail.

However, we will now require Get Message Type Support to be implemented; it is mandatory according to DSP0236. Bump the warning for this to non-verbose mode too.

This means we need to adjust the properties-query-retry tests, to expect enumeration failure on the max retries case.

While we're here, make the error paths consistent between the two callers of query_peer_properties.

Fixes: https://github.com/CodeConstruct/mctp/issues/161

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
    invalid, having a zeroed control message header. These are now properly
    populated
 
+3. `mctpd` will no longer (incorrectly) reject SetupEndpoint / AssignEndpoint
+   calls for endpoints that do not implement the optional Get Endpoint UUID
+   and Get Vendor Defined Message Support commands.
+
 ### Changed
 
 1. `mctpd`'s `mode` configuration (setting bus owner vs. endpoint roles) is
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
    invalid, having a zeroed control message header. These are now properly
    populated
 
+3. `mctpd` will no longer (incorrectly) reject SetupEndpoint / AssignEndpoint
```
msnidhin, line 34: AssignEndpointStatic also will be successful right?

jk-ozlabs: Yep, that's all the same path.

msnidhin:
Hi @jk-ozlabs  Is it possible to merge this ?

PR #163, opened 2026-07-06
Update mctpd and example configs to minimise necessary permissions

SafetyInObscurity:
Updated mctpd to manage its capabilities, dropping `CAP_NET_BIND_SERVICE` when no longer needed. The example configs have been updated to run mctpd as an unprivileged user with only necessary capabilities added while also allowing it to continue to be run as root if desired. 

The ability to interact with mctpd via the system D-bus has also been restricted to root and a special `mctp-admin` role, though properties can still be viewed by unprivileged users.

### DIFF
--- a/docs/mctpd.md
+++ b/docs/mctpd.md
@@ -503,3 +503,14 @@ match = { path = "/devices/pci0000:00/0000:00:08.3/*" }
 ```
 
 Paths have the `/sys` prefix stripped.
+
+## Capabilities
+
+`mctpd` requires the `CAP_NET_BIND_SERVICE` and `CAP_NET_ADMIN` capabilites to run.
+If the service is not running as root these should be added to the `AmbientCapabilities`
+setting in the service's config.
+If the target platform has the `libcap` library available,
+the `mctpd` process will drop `CAP_NET_BIND_SERVICE` when it is no longer needed.
+
+Because `mctpd` may be exposed to potentially hostile inputs,
+it shouldn't be given more capabilities than necessary.
--- a/meson.build
+++ b/meson.build
@@ -16,6 +16,8 @@ add_project_arguments('-Wno-unused-parameter', language : 'c')
 
 libsystemd = dependency('libsystemd', version: '>=247', required: false)
 
+libcap = dependency('libcap', required: false)
+
 conf = configuration_data()
 conf.set10('HAVE_LINUX_MCTP_H',
     cc.has_header('linux/mctp.h'),
@@ -36,6 +38,10 @@ conf.set10('MCTPD_RECOVER_NIL_UUID',
 conf.set10('MCTPD_WRITABLE_CONNECTIVITY',
     get_option('unsafe-writable-connectivity'),
     description: 'Allow writes to the Connectivity member of the au.com.codeconstruct.MCTP.Endpoint1 interface on endpoint objects')
+
+conf.set10('HAVE_LIBCAP',
+    libcap.found(),
+    description: 'Require mctpd to minimise its capabilities')
 
 conf.set_quoted('MCTPD_CONF_FILE_DEFAULT',
     join_paths(get_option('prefix'), get_option('sysconfdir'), 'mctpd.conf'),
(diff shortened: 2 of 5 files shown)


docs/mctpd.md
```diff
@@ -503,3 +503,13 @@ match = { path = "/devices/pci0000:00/0000:00:08.3/*" }
 ```
 
 Paths have the `/sys` prefix stripped.
+
+## Capabilities
+
+`mctpd` requires the `CAP_NET_BIND_SERVICE` and `CAP_NET_ADMIN` capabilites to run.
+If the service is not running as root these should be added to the `AmbientCapabilities` setting in the service's config.
```
jk-ozlabs: To be consistent with the rest of the file, wrap this at 80 cols.

src/mctpd.c
```diff
@@ -6342,6 +6346,50 @@ static int endpoint_allocate_eids(struct peer *peer)
 	return 0;
 }
 
+#if MCTPD_MANAGE_CAPABILITIES
+
+int drop_bind_cap(void)
```
jk-ozlabs: Should be `static`, in both this and the no-op implementation below.

meson.build
```diff
@@ -100,7 +106,7 @@ if libsystemd.found()
             'src/mctpd.c',
         ] + test_ops_sources + netlink_sources + util_sources,
         include_directories:  include_directories('src'),
-        dependencies: [libsystemd, toml_dep],
+        dependencies: [libsystemd, toml_dep, libcap],
```
jk-ozlabs, line 109: it might be nice to prevent the test-mctpd process from attempting to alter capabilities - its `ops->bind()` doesn't really do a `bind()`, as it's a mock implementation, and so it shouldn't have any capabilities in the first place. We could do the caps change as a separate op, but that seems pretty heavy-handed.

Any ideas to handle that?

[I am assuming that the absence of `CAP_NET_BIND_SERVICE` in the first place should not be an issue for the drop.]

meson.build
```diff
@@ -37,6 +39,10 @@ conf.set10('MCTPD_WRITABLE_CONNECTIVITY',
     get_option('unsafe-writable-connectivity'),
     description: 'Allow writes to the Connectivity member of the au.com.codeconstruct.MCTP.Endpoint1 interface on endpoint objects')
 
+conf.set10('MCTPD_MANAGE_CAPABILITIES',
```
jk-ozlabs: Slight preference to name this according to the dependency it is representing (ie., `HAVE_LIBCAP`), rather than as a user-provided option - as these last two `MCTPD_` conf parameters are doing.

meson.build
```diff
@@ -37,6 +39,10 @@ conf.set10('MCTPD_WRITABLE_CONNECTIVITY',
...
+    libcap.found(),
+    description: 'Require mctpd to minimise its capabilities (emits warning on start if false)')
```
jk-ozlabs: This looks stale, as no warning is generated. I'm happy for there to be no warning, but we should make this consistent.

jk-ozlabs (commented):
Looks good! A couple of minor comments inline, and a bit of an open-ended query.

meson.build
SafetyInObscurity: I don't think there should be any trouble from trying to set the capabilities to their current value.

If you really wanted to prevent it from trying to change capabilities during testing, there should be ways to have `MCTPD_MANAGE_CAPABILITIES` set to `0` for the test.

Is there anything specific you're worried about?

meson.build
SafetyInObscurity: I was thinking of what the macro meant for the code's behaviour, rather than what conditions allowed it.

Figured some users might not want to include the functionality even if libcap was available.

I probably should have added it as a configurable option.

jk-ozlabs: No need to make it a config option; if libcap is available we should use it.

Although it's possible to provide an option for things like this, it's not always desirable. In this case, there's a fairly obvious ideal choice, so we take that. If someone has a use-case for something else, we can incorporate that into a future change. Since we don't have details of that use case now though, no need to cater for it.

PR #164, opened 2026-07-14
Correct documentation and update mctp tool help

SafetyInObscurity:
There are some minor errors, inconsistencies, and duplications in the documentation and `mctp` tool help. I've aimed to make common-sense corrections.

### DIFF
--- a/src/mctp.c
+++ b/src/mctp.c
@@ -895,10 +895,10 @@ static int cmd_addr(struct ctx *ctx, int argc, const char **argv)
 	const char *subcmd;
 	if (argc == 2 && !strcmp(argv[1], "help")) {
 		fprintf(stderr, "%s address\n", ctx->top_cmd);
-		fprintf(stderr, "%s address show [IFNAME]\n", ctx->top_cmd);
-		fprintf(stderr, "%s address add <eid> dev <IFNAME>\n",
+		fprintf(stderr, "%s address show [ifname]\n", ctx->top_cmd);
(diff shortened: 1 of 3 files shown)


src/mctp.c
```diff
@@ -1126,12 +1126,12 @@ static int cmd_route(struct ctx *ctx, int argc, const char **argv)
 		fprintf(stderr, "%s route show [net <network>]\n",
 			ctx->top_cmd);
 		fprintf(stderr,
-			"%s route add <eid>[-<eid>] via <dev> [mtu <mtu>]\n",
+			"%s route add <eid>[-<eid>] via <IFNAME> [mtu <mtu>]\n",
```
mkj: Makes sense. Lower case might be more consistent with the rest of the args? (There are others upper case to change too)

jk-ozlabs:
Patches look good! 1/2 makes sense, and the consistency in 2/2 is definitely an improvement. One thing we may need to decide on though:

```
mctp [...] dev <dev>
```

vs.

```
mctp [...] dev <ifname>
```

`ifname` is definitely more descriptive, but `dev` is the literal we use in the command, so may have a lower cognitive load to recall. (ie., "is it `dev <ifname>` or `ifname <dev>` ?"). I think I am fine with either, but we should be intentional in choosing.

SafetyInObscurity:
I think it's better to use `dev <ifname>` since it gives the users the extra information that it's the name of an interface, but `dev <dev>` doesn't tell them anything they can't guess. It also helps distinguish from `mctp link serial <device>` which is an actual device file.

PR #168, opened 2026-08-06
mctp-client: add cci message type code

LeoYangQuanta:
According to Specification DSP0239 1.12.0
Add type code 8 for use with the CXL Type 3 Device Component Command Interface over MCTP.

Testing
---
cci commands
```
mctp-client net 1 eid 10 type cci data 00 00 00 00 03 00 00 00 00 00 cd ab
01 00 00 00 03 08 00 00 00 00 00 00 2B 01 AD 9C A4 8A D2 16
```

### DIFF
--- a/src/mctp-client.c
+++ b/src/mctp-client.c
@@ -67,6 +67,12 @@ static struct type_lookup_t {
 			"secured messages using spdm over mctp per DSP0276",
 	},
 	{
+		.name = "cxl-cci",
+		.type = 8,
+		.description =
+			"CXL type 3 device component command interface over mctp per DSP0281",
+	},
+	{
 		.name = "pci",
 		.type = 0x7e,
 		.description = "vdm using a pci based vendor id per DSP0236",


src/mctp-client.c
```diff
@@ -66,6 +66,12 @@ static struct type_lookup_t {
 		.description =
 			"secured messages using spdm over mctp per DSP0276",
 	},
+	{
+		.name = "cci",
```
jk-ozlabs: should we clarify this is CXL related? perhaps `cxl-cci`?

jk-ozlabs (changes requested):
Change itself looks okay, just a comment on the chosen name here.

However, the commit message just needs a little work. The wording is a little off, and no need for the section break there, nor that `Testing` line (we assume that you have tested this before submission).

Maybe just:

> Add type code 8 for CXL Type 3 Device Component Command Interface,
>  as defined by DSP0239 1.9.0 and later.
> 
> For example:
> ```
> $ mctp-client eid 10 type cxl-cci data 00 00 00 00 03 00 00 00 00 00 cd ab
> 01 00 00 00 03 08 00 00 00 00 00 00 2B 01 AD 9C A4 8A D2 16
> ```

src/mctp-client.c
LeoYangQuanta: Yes, that's more clear. Thank you.

jk-ozlabs (approved):
Looks good, thanks!

PR #171, opened 2026-08-18
headers: update for compatibility with toolchain headers down to 5.15

SafetyInObscurity:
Found that many distro build toolchains use outdated kernel headers. Explicitly add linux/if.h to mctp-ops-test.c, and add missing linux/mctp.h definitions if undefined.

### DIFF
--- a/src/mctp.h
+++ b/src/mctp.h
@@ -109,4 +109,28 @@ enum {
 #define RTNLGRP_MCTP_IFADDR 34
 #endif
 
+#ifndef SIOCMCTPALLOCTAG
+#define SIOCMCTPALLOCTAG (SIOCPROTOPRIVATE + 0)
+#define SIOCMCTPDROPTAG (SIOCPROTOPRIVATE + 1)
+
+struct mctp_ioc_tag_ctl {
+	mctp_eid_t peer_addr;
+	__u8 tag;
+	__u16 flags;
+};
+#endif
+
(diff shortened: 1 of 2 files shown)


src/mctp.h
```diff
@@ -109,4 +109,28 @@ enum {
 #define RTNLGRP_MCTP_IFADDR 34
 #endif
 
+#ifndef SIOCMCTPALLOCTAG
+#define SIOCMCTPALLOCTAG 0x89E0
```
jk-ozlabs: I think we can rely on `SIOCPROTOPRIVATE` being valid for the versions we care about.

jk-ozlabs (changes requested):
Looks good, I'd just suggest a tweak to remove the hard-coded `SIOCPROTOPRIVATE` value.

src/mctp.h
SafetyInObscurity: Done

jk-ozlabs: The new version seems to omit the `SIOCMCTPALLOCTAG` definition, which we do need.

The comment above was referring to using the `SIOCPROTOPRIVATE` rather than the hard-coded `0x89e0`

SafetyInObscurity: Ah, that's what happens when I don't read it
One sec.

SafetyInObscurity: Done for real this time

jk-ozlabs:
Merged, thanks!

PR #172, opened 2026-08-20
build: add tags for packaging

SafetyInObscurity:
Add tags to mctp utilities, mctpd, and debug tools to allow packagers to easily select features. Tag debug tools as installable to allow them to be installed when needed by packagers.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,7 @@
 #
 project(
     'mctp', 'c',
-    meson_version: '>= 0.59.0',
+    meson_version: '>= 0.60.0',
     version: 'v2.6',
     license: 'GPLv2',
     default_options: [
@@ -66,6 +66,7 @@ toml_dep = declare_dependency(
 executable('mctp',
     sources: ['src/mctp.c'] + netlink_sources + util_sources + ops_sources,
     install: true,
+    install_tag: 'utils',
 )
 
 mctp_test = executable('test-mctp',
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
The `install_tags` support looks interesting, but can you elaborate on the intended use-case here?

Most package-build infrastructure should be able to specify what upstream-package-installed binaries actually get included into the distribution package, so what do the tags give us above that?

Also, this will require a bump to the minimum meson version, I think.

SafetyInObscurity:
> The `install_tags` support looks interesting, but can you elaborate on the intended use-case here?
>
> Most package-build infrastructure should be able to specify what upstream-package-installed binaries actually get included into the distribution package, so what do the tags give us above that?

It simplifies the installation process for package-build infrastructure. Yocto and Buildroot for example have specific Meson integrations that automatically install binaries tagged with `install : true`. Without `install_tag` it's all or nothing, all installable binaries get added. Unwanted binaries have to be manually excluded or uninstalled, and absent binaries need to be added. 

This project is broadly the mctp utils, mctpd, and the debug tools. The different groups don't depend on each other and any combination of them might be wanted, so I thought tagging them that way would be helpful.

It also helps anyone who wants to automatically install a subset of the binaries.

> Also, this will require a bump to the minimum meson version, I think.

I agree, installation tags were added in 0.60.0, so the minimum version would need bumping.

SafetyInObscurity:
Added the meson version bump to the PR

Issue #149, opened 2026-03-30
Support for `Recover` method on bridged endpoints


jk-ozlabs:
Currently, a `Recover` invocation will attempt a Get Endpoint ID (to check endpoint connectivity), but will do so using a physical addressing:

```c
	/*
	 * Test if we still have connectivity to the endpoint. If we do, we will get a
	 * response reporting the current EID. This is the test recommended by 8.17.6
	 * of DSP0236 v1.3.1.
	 */
	rc = query_get_endpoint_id(ctx, &peer->phys, &peer->recovery.eid,
				   &peer->recovery.endpoint_type,
				   &peer->recovery.medium_spec, /*peer=*/NULL);
```

The `NULL`  peer causes the physical addressing there.

Since we cannot target bridged endpoints with a phys address, we will need to use vanilla EID addressing for those.

faizana-nvidia:
Hello @jk-ozlabs 

I too was looking for this, if you still haven't picked it up as your current working item, I'll be happy to push a change.

jk-ozlabs:
@faizana-nvidia sure, if you want to look at this!

I'd suggest working out a plan here before writing the implementation though, there are some design points for the recovery flow that we would need to sort out too. Mainly:

 * we'll probably want to check connectivity to the bridge too
 * we'll need to handle cases where the EID has been reassigned by the downstream bus owner (ie, has a new UUID), by generating a remove & re-add.

faizana-nvidia:
great! let me update here with the flow

faizana-nvidia:
# Downstream Endpoint Recovery Flow

## method_endpoint_recover (D-Bus entry point) — unchanged for both peer types

```mermaid
flowchart TD
    A([method_endpoint_recover called]) --> B{peer->degraded?}
    B -- yes --> Z([no-op, return OK])
    B -- no --> C["set recovery.npolls, recovery.delay\nschedule peer_endpoint_recover timer\npeer->degraded = true\nemit Connectivity changed"]
    C --> D([return — common for direct and downstream peers])
```

---

## peer_endpoint_recover (timer callback) — downstream branch inserted after physical probe

```mermaid
flowchart TD
    TIMER([timer fires: peer_endpoint_recover]) --> DEC[Decrement recovery.npolls]

[...]

PR #175, opened 2026-08-24
Fix downstream endpoint recovery

faizana-nvidia:
This change introduces mechanism to trigger recovery for endpoints which are behind the bridge. Current Recovery dbu-method doesn't work if triggered for downstream endpoint due to lack to right query path (addressed to downstream endpoint) with this PR we aim to merge with existing logic while handling downstream endpoint use case as well. Originally issue was being tracked via [#149](https://github.com/CodeConstruct/mctp/issues/149)

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -2954,7 +2954,7 @@ out:
 	return rc;
 }
 
-static int query_get_peer_uuid(struct peer *peer)
+static int query_get_peer_uuid(struct peer *peer, uint8_t uuid[16])
 {
 	struct mctp_ctrl_resp_get_uuid *resp = NULL;
 	struct mctp_ctrl_cmd_get_uuid req;
@@ -2985,10 +2985,7 @@ static int query_get_peer_uuid(struct peer *peer)
 		goto out;
 
 	resp = cmd.resp;
-	rc = peer_set_uuid(peer, resp->uuid);
-	if (rc < 0)
-		goto out;
-	rc = 0;
+	memcpy(uuid, resp->uuid, sizeof(resp->uuid));
 
 out:
 	mctp_ctrl_cmd_free(&cmd);
@@ -3458,10 +3455,19 @@ static int query_peer_properties(struct peer *peer)
 		}
 	}
 
-	rc = query_get_peer_uuid(peer);
-	if (rc < 0 && peer->ctx->verbose) {
-		errno = -rc;
-		warn("Error getting UUID for %s", peer_tostr(peer));
+	{
+		uint8_t uuid[16];
+		memset(uuid, 0, sizeof(uuid));
+		rc = query_get_peer_uuid(peer, uuid);
+		if (rc < 0) {
+			if (peer->ctx->verbose) {
+				errno = -rc;
+				warn("Error getting UUID for %s",
+				     peer_tostr(peer));
+			}
+		} else {
+			rc = peer_set_uuid(peer, uuid);
+		}
 	}
 
 	// TODO: emit property changed? Though currently they are all const.
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -224,6 +224,9 @@ struct peer {
 	uint8_t pool_size;
 	uint8_t pool_start;
 
+	// Next hop toward the bus owner for downstream peer
+	struct peer *gateway;
+
 	struct {
 		sd_event_source **sources;
 	} bridge_ep_poll;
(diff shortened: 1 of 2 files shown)


jk-ozlabs:
Thanks for the PR. One main thing to start:

>   1. Physical liveness check: send Get Endpoint ID to the bridge's
>     hardware address.
>  2. EID-routed probe: send Get Endpoint ID via EID routing through the
>     bridge to verify the downstream device is still present.
>  3. UUID check via EID routing: compare the UUID returned by the
>     downstream device against the stored UUID to detect device exchange.
>     A UUID mismatch triggers InterfacesRemoved + InterfacesAdded so the
>     new device is registered cleanly.

This seems a bit backwards - I would have expected that we perform the Get Endpoint ID query to the endpoint first, to establish base connectivity. If that fails, then walk up the tree to find the top-most bridge with the connectivity issue.

Is there a reason to interact with the bridge _first_?

faizana-nvidia:
> Is there a reason to interact with the bridge first?

The way I was looking into this was, since everything will be routed via bridge if bridge itself is non responsive then probably downstream path might be broken too and we could be seeing timeouts with probe on downstream EID.

jk-ozlabs:
> The way I was looking into this was, since everything will be routed via bridge if bridge itself is non responsive then probably downstream path might be broken too and we could be seeing timeouts with probe on downstream EID.

Yes, I understand the rationale for checking connectivity to the bridge, but I think the ordering should be reversed - that we want to check for the smallest failure first - being the actual endpoint. If we have lost connectivity there, *then* we can walk up the bridge topology. Keep in mind there may be multiple bridges to walk!

Also: a call to `Recover` may not mean we have actually lost connectivity to the downstream endpoint at all, just that something has gone wrong at the application level, and that mctpd should query it.

faizana-nvidia:
Ack, I'll update the logic to change the order. Also when you mention about multiple bridge walk, I do get the topological aspect and wanted to find out which bridge comms exactly broke, but cascaded bridge system support is not yet added, so may be we can delegate it for later?

jk-ozlabs:
> but cascaded bridge system support is not yet added, so may be we can delegate it for later?

Yes, but you don't need to delegate much - at present, it will just appear as though the top-level bridge of a cascade just hosts all downstream devices (ie,. we do not have visibility on the intermediate bridges with the current code).

So, once we have a concept of those intermediate bridges, as long as the recovery code has a "query the peer's bridge" path, we would just change that to iterate over the tree (ie, to the peer's bridge's bridge, etc) as suitable.

However: with the intermediate bridges missing from the topology, we may see the following case:

```mermaid
graph TD;
    TMBO-->B1[Bridge 1];
    B1--> B2[Bridge 2];
    B2--> EP;
```

If bridge 2 is offline, the EP will be unreachable, but the only bridge that the TMBO is aware of is Bridge 1, which will appear to have full connectivity. We'll need to handle that case gracefully.

faizana-nvidia:
Ack, I see two ways of doing this.

1. Assuming our peer->pool_start and peer->pool_end is maintained in our code down to every bridge in the path.
Bridge1 [PS1, PE1], Bridge2[PS2, PE2],..

Although I'm not sure how could this be maintained since there is probably no direct way of asking bridge about its allocation results (Pool Start, Pool End) apart from what TBO knows about its nearest Bridge, but if this is somehow maintained then :

We can recursively trace each bridges which might contain the endpoint eid E i.e (PSx < E < PEx)  from TBOs nearest and check bridge's connectivity till we reach the last bridge nearest to Endpoint E. If intermediate bridge is point of failure then simply flag that Bridge EID as break point, may be?

2. Introduce Query Hop may be to systematically ask for nearest Bridge for given endpoint EID and then check connectivity to that bridge eid, all the way down till query hop successfully returns next bridge eids

jk-ozlabs:
Yes, for future features, we'll need more awareness of the downstream bridge layout and EID allocations, and I figure the recovery paths would just consume that. That would likely result in maintaining a tree data structure (say, maintainting `peer->gateway`) rather than having to scan through EID pool ranges.

We could look at either query hop or GRTE to populate that, I guess.

faizana-nvidia:
> Yes, for future features, we'll need more awareness of the downstream bridge layout and EID allocations, and I figure the recovery paths would just consume that. That would likely result in maintaining a tree data structure (say, maintainting `peer->gateway`) rather than having to scan through EID pool ranges.
> 
> We could look at either query hop or GRTE to populate that, I guess.

I guess I understand the approach here, for GRTE capable bridges its direct approach of updating `gateway` but for polling based discovery (for non GRTE capable bridges) we  probably would need Query HOP, which itself is not a mandatory command.

Also would it be okay if i introduce Query HOP logic in this PR as part of polling discovery?

jk-ozlabs:
I don't think we want to introduce the full topology implementation in this change.

faizana-nvidia:
I've updated the patch with rework, its open for review

src/mctpd.c
```diff
@@ -2954,7 +2957,7 @@ static int query_get_peer_uuid_by_phys(struct ctx *ctx, const dest_phys *dest,
 	return rc;
 }
 
-static int query_get_peer_uuid(struct peer *peer)
+static int query_get_peer_uuid(struct peer *peer, uint8_t uuid[16])
```
jk-ozlabs, line 2960: Minor, but splitting this out to a prereq patch (in the same series, of course) may make the actual changes more clear. Up to you on whether that's worthwhile.

src/mctpd.c
```diff
@@ -3751,35 +3760,47 @@ static int peer_endpoint_recover(sd_event_source *s, uint64_t usec,
...
-		goto reschedule;
-	}
+	if (peer->gateway) {
+		/*
+		 * For downstream endpoint recovery :
+		 *
+		 * If EID probe fails, check if the bridges are still responsive by walking up the bridge chain diagnosting point of exchange.
+		 * In case we've got a response, verify UUID to detect device exchange at the same pool EID.
+		 * for non matching UUID, remove old peer and publish the new peer with the same EID.
+		 */
```
jk-ozlabs, line 3771: Minor things:

 - please keep the wrapping consistent with the rest of the file
 - `diagnosting` → `diagnosting`

not clear on what "point of exchange" is supposed to indicate though, perhaps "diagnosing loss of connectivity" ?

jk-ozlabs (commented):
I think this looks pretty good. Just a couple of minor things inline.

@amboar -  since the original recovery implementation is yours, can you take a look too?

src/mctpd.c
faizana-nvidia: Ack, will do the wrapping.

> not clear on what "point of exchange" is supposed to indicate though, perhaps "diagnosing loss of connectivity" ?

yeah this sounds better, I'll update

src/mctpd.c
faizana-nvidia: I can add a new commit to capture this change separately.

faizana-nvidia:
Addressed the comments

src/mctpd.c
```diff
@@ -2985,10 +2988,7 @@ static int query_get_peer_uuid(struct peer *peer)
 		goto out;
 
 	resp = cmd.resp;
-	rc = peer_set_uuid(peer, resp->uuid);
-	if (rc < 0)
-		goto out;
-	rc = 0;
+	memcpy(uuid, resp->uuid, 16);
```
amboar: `sizeof(uuid)`?

jk-ozlabs: there's a few instances of a magic `16` around. Is there a better way to represent the UUID types perhaps?

amboar: I think that would be a good, though I didn't leave a request to that end as I didn't want to lump @faizana-nvidia with refactorings that weren't entirely their concern.

That said, I doubt it's a lot of trouble to address.

jk-ozlabs: Yeah, let's keep it simple here; avoid adding new constants where possible, but can defer wider cleanups to a separate change.

faizana-nvidia: > Yeah, let's keep it simple here; avoid adding new constants where possible, but can defer wider cleanups to a separate change.

okay its minor change, I'll quickly address this. Thanks

faizana-nvidia:
gentle reminder.

jk-ozlabs:
> gentle reminder.

have some other fixes in progress at the moment, will take a look at this when that's sorted.

src/mctpd.c
```diff
@@ -3757,35 +3760,48 @@ static int peer_endpoint_recover(sd_event_source *s, uint64_t usec,
...
+					ctx, &gw->phys, &peer->recovery.eid,
+					&peer->recovery.endpoint_type,
+					&peer->recovery.medium_spec,
+					gw->gateway ? gw : NULL,
+					/*retry=*/false);
+				if (gw_rc < 0)
+					warnx("Recovery: EID %d unreachable; bridge EID %d failed to respond",
+					      peer->eid, gw->eid);
+				gw = gw->gateway;
+			}
```
jk-ozlabs, line 3796: I figure we're OK with the multiple commands here, as the gateway tree should be fairly flat, we don't retry, and we abort on the first failure, so won't block dbus operation for extended periods.

A couple of things to confirm though:
 - we're OK with the not-retrying, in case of transient errors; and
 - we're OK with not having handling for the peer -> gateway becoming cyclic

The latter would definitely be a bug, and I assume we are OK with not handling that case here.

faizana-nvidia: > we're OK with the not-retrying, in case of transient errors; and

I suppose we are rescheduling the whole process recovery flow again which would reattempt on the gateway chain thus covering transient errors.

> we're OK with not having handling for the peer -> gateway becoming cyclic

yeah this is something I hadn't given really a though as I assumed flat structure like you mentioned, but sure we can pick this case for later may be?

jk-ozlabs: > I suppose we are rescheduling the whole process recovery flow again

Ah yeah, good point.

> as I assumed flat structure like you mentioned, but sure we can pick this case for later may be?

Yeah, we require a flat structure; we don't need to support cycles in the routing graph as a working case.

However, should we be checking that the graph is OK, with the introduction of the gateway pointers?

faizana-nvidia: > However, should we be checking that the graph is OK, with the introduction of the gateway pointers?

I suppose we can introduce a check to figure out if we are looping while traversing the gateway chain during recovery. If that does happen, then what would be the fallback here? can't recover the endpoint? 

Or should we be rather worry about the graph while peers are getting discovered and we are adding gateway?

PR #176, opened 2026-08-28
mctpd: handle Discovery Notify (DSP0236 command 0x0D)

FXN-Bruce:
mctpd currently falls through to handle_control_unsupported() for an
incoming Discovery Notify, replying with an unsupported-command error
and taking no other action.

Add a handler that acknowledges the message (using physical
addressing, since the sender may not have a valid EID yet) and emits
a "DiscoveryNotify" D-Bus signal on the sender's interface object
(au.com.codeconstruct.MCTP.BusOwner1 at
/au/com/codeconstruct/mctp1/interfaces/<ifname>). The handler does not
otherwise inspect or modify peer/EID state itself - reconciling what
a Discovery Notify implies for a given peer is bus-owner policy, left
to whatever subscribes to the signal.

### DIFF
--- a/src/mctpd.c
+++ b/src/mctpd.c
@@ -1330,6 +1330,61 @@ handle_control_endpoint_discovery(struct ctx *ctx, int sd,
 	return reply_message_phys(ctx, sd, resp, sizeof(*resp), addr);
 }
 
+/* Handles an incoming Discovery Notify (DSP0236 command 0x0D): a device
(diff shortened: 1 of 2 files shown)


jk-ozlabs:
Please sync with #165.

FXN-Bruce:
> Please sync with #165.

I see. Thank you. Withdrawing.

PR #179, opened 2026-09-07
mctpd: grant CAP_NET_RAW

msnidhin:
The kernel MCTP socket implementation requires CAP_NET_RAW to send MCTP control messages when mctpd runs as an unprivileged user.

### DIFF
--- a/conf/mctpd.service
+++ b/conf/mctpd.service
@@ -9,7 +9,7 @@ Type=dbus
 BusName=au.com.codeconstruct.MCTP1
 ExecStart=/usr/sbin/mctpd
 User=mctpd
-AmbientCapabilities=CAP_NET_BIND_SERVICE CAP_NET_ADMIN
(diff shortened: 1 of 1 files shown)


msnidhin:
Hi.. During testing of latest mctpd observed one issue
mctpd[278]: mctpd: endpoint_query_addr: sendto(sockaddr_mctp_ext eid XX net X type 0x00 if 0 hw len 0 0x) 2 bytes failed. Permission denied

@jk-ozlabs  @SafetyInObscurity  
Please review if this is the correct fix

jk-ozlabs:
Looks good. We may want to relax the `CAP_NET_RAW` requirement in general though, but this seems like a reasonable interim measure.

@SafetyInObscurity , any thoughts?

SafetyInObscurity:
Looks like the right fix. An oversight on my part. Will probably need to bump the version to 2.7 after merging this to fix upstream distros

SafetyInObscurity:
Have confirmed the error and that the change resolves it

jk-ozlabs:
Merged, thanks!

